PR #3, opened 2022-02-08
Feat: Improved Node runtime

Meldiron:
- Renamed from `example.sh` to `deploy.sh`, as it is more fitting name
- Simplified example usage (better developer experience)
- Made example usage more useful (outside of the open-runtimes repo)
- Added default values for `path` and `file`, so developer can omit them if they put code in `index.js`
- Added default response if function has no response
- Improved error catching to support both sync and async functions


runtimes/node-17.0/server.js
```diff
@@ -15,22 +15,33 @@ const server = micro(async (req, res) => {
...
+        },
+        json: (json, status = 200) => {
+            isResponded = true;
+            send(res, status, json);
+        },
     };
+
     try {
-        let userFunction = require(path.join(body.path, body.file));
+        let userFunction = require(path.join(body.path ?? '/usr/code', body.file ?? 'index.js'));
```
Meldiron, line 31: Default should be CONST

runtimes/node-17.0/README.md
```diff
@@ -59,3 +86,7 @@ module.exports = (req, res) => {
     });
 }
 ```
+
+- To handle dependencies, you need to have `package.json` file. Dependencies will be automatically cached and installed, so you don't need to include `node_modules` folder in your function.
```
eldadfux, line 90: Is the runtime responsible for installing the dependencies? Don't we delegate this to the builder in Appwrite? Meaning the runtime should just get everything ready for execution?

README.md
```diff
@@ -63,21 +64,24 @@ All runtimes share a common basic structure, but each additionally adds runtime-
...
+| example/           	| Contains a sample function to demonstrate the usage of the runtime server                                                                             	|
+| docker-compose.yml 	| Configuration to easily run the example code with `docker-compose up`                                                                                 	|
+| Dockerfile         	| Instructions to build a runtime, install it's dependencies and setup the runtime server. These images are usually based on official alpine or ubuntu. 	|
+| server.X           	| A HTTP server implemented in the respective runtime's language. File extension depends on your runtime. For instance, Python is `server.py`           	|
+| build.sh           	| Script responsible for building user code. This can be package installations, or any specific build process the runtime requires.                     	|
+| launch.sh          	| Script to launch the HTTP server on port `3000`. Additionally, it also copies the user supplied code to a directory accessible to the server.         	|
```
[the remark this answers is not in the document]
Meldiron: `launch` takes `.tar.gz` file as input, extracts it and prepares it to be run (for example in node, it copies node_modules to HTTP server script, because that is now Node loads dependencies).

The last step of `launch` is `npm start` which runs the `server.js`.

README.md
```diff
@@ -16,7 +16,7 @@ Runtime environments for serverless cloud computing for multiple coding language
   - [Features](#features)
   - [Roadmap](#roadmap)
   - [Images](#images)
-  - [Runtimes introduction](#runtimes-introduction)
+  - [Repository Structure](#repository-structure)
```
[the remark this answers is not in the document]
Meldiron: Moved below

runtimes/node-17.0/README.md
Meldiron: Runtime consists of multiple scripts. One of them is `build`, one is `launch`.

Build gets package.json, and prepares folder that is ready to be executed.

Launch takes this folder (gzipped version of it) and launches it.

Also there is new `deploy.sh`, which takes these two actions and merges it. With deploy, you can point it to your code (index.js and package.json), and it will first build it (generating node_modules folder), and then run it (HTTP server).

---

Does the description make sense now, or should we update it?

runtimes/node-17.0/README.md
```diff
@@ -59,3 +86,7 @@ module.exports = (req, res) => {
...
+
+- The default entrypoint is `index.js`. If your entrypoint differs, make sure to provide it in the JSON body of the request: `{"file":"src/app.js"}`.
```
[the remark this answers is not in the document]
Meldiron: Added. We don't yet have a contribution guide, so I linked to broken file right now.

runtimes/node-17.0/README.md
```diff
@@ -59,3 +86,29 @@ module.exports = (req, res) => {
...
+
++ [https://github.com/PineappleIOnic](https://github.com/PineappleIOnic)
+
+**Matej Bačo**
+
++ [https://github.com/Meldiron](https://github.com/Meldiron)
+
+## Contributing
+
+For security issues, please email security@appwrite.io instead of posting a public issue in GitHub.
```
eldadfux, line 112: Let's have a separate title for `Security`.

eldadfux (approved):

PR #9, opened 2022-02-15
Implemented tests for runtimes

Meldiron:


tests/node-17.0/tests.js
```diff
@@ -0,0 +1,25 @@
...
+    If an error is thrown, a response with code 500 will be returned.
+*/
+
+module.exports = async (req, res) => {
+    const payload = JSON.parse(req.payload);
+
+    const todo = await fetch(`https://jsonplaceholder.typicode.com/todos/${payload.id ?? 1}`).then(r => r.json());
+
+    res.json({
+        isTest: true,
```
eldadfux, line 21: We might want to add some env vars and user input so we can validate they are being accepted by the function as expected.

tests/node-17.0/tests.js
```diff
@@ -0,0 +1,25 @@
...
+        'send(text, status)' - function to return text response. Status code defaults to 200
+        'json(obj, status)' - function to return JSON response. Status code defaults to 200
+    
+    If an error is thrown, a response with code 500 will be returned.
+*/
+
+module.exports = async (req, res) => {
+    const payload = JSON.parse(req.payload);
+
+    const todo = await fetch(`https://jsonplaceholder.typicode.com/todos/${payload.id ?? 1}`).then(r => r.json());
```
eldadfux, line 18: I never feel comfortable with 3rd party APIs. How important is this text? will we get the same result from just checking a simple computation?

tests/Node170.php
```diff
@@ -0,0 +1,67 @@
...
+
+use PHPUnit\Framework\TestCase;
+
+ini_set('display_errors', 1);
+ini_set('display_startup_errors', 1);
+error_reporting(E_ALL);
+
+class Node170 extends TestCase
+{
+    private function call($body) {
```
eldadfux, line 13: This method could be inherited from an abstract test class used for all languages. We can also have all languages tests on the same PHPUnit class with different methods..

tests/node-17.0/tests.js
Meldiron: We use ID of todo from payload, input should be tested. Ill add tests for environment variables and headers.

tests/node-17.0/tests.js
Meldiron: We do http request so we have standard library test we could do in all languages. We could use our own api, but I would strongly recommend testing runtime with a http request, because it tests if libraries work in a really simple, easy to understand and language agnostic way.

tests/Node170.php
Meldiron: I see. Functions (input and output) are same, just the implementation is different. Ill make sure to abstract it.

tests/node-17.0/tests.js
eldadfux: 👌

tests/node-17.0/tests.js
eldadfux: OK 👌

tests/Node170.php
eldadfux: 👌

tests/Base.php
```diff
@@ -0,0 +1,92 @@
...
+
+use PHPUnit\Framework\TestCase;
+
+ini_set('display_errors', 1);
+ini_set('display_startup_errors', 1);
+error_reporting(E_ALL);
+
+abstract class Base extends TestCase
+{
+    protected string $endpoint;
```
eldadfux, line 13: Would entrypoint be a better name?

README.md
```diff
@@ -132,6 +132,24 @@ All requests should also have JSON body with the following structure:
 
 All body parameters are optional. The values used in the example above are the default values.
 
+## Testing
```
christyjacob4, line 135: Please add this to the contents section

christyjacob4 (changes requested):
Just some minor comments

christyjacob4 (approved):

PR #10, opened 2022-02-17
Feat: Dart runtime

lohanidamodar:


lohanidamodar:
@Meldiron @PineappleIOnic I removed the `function_types` package and created it as a part of server for now, as @Meldiron  pointed that we are not going to release extra packages for first release, if we will not be releasing `function_types` package, we cannot use it in examples. As devs will have no way of getting that package. @Matej Bačo pointed for other strictly typed language as well we are not releasing the types package.

PR #11, opened 2022-02-17
Feat: PHP runtime

Meldiron:


PineappleIOnic (approved):

runtimes/php-8.0/composer.json.fallback
```diff
@@ -0,0 +1,17 @@
+{
+    "name": "user/function",
+    "description": "",
+    "type": "library",
+    "license": "MIT",
```
eldadfux, line 5: What is this file used for?

runtimes/php-8.0/prepare.php
```diff
@@ -0,0 +1,4 @@
+<?php
```
eldadfux, line 1: What is this file used for?

runtimes/php-8.0/server.php
```diff
@@ -0,0 +1,86 @@
...
+    $request = [
+        'env' => $body['env'] ?? [],
+        'headers' => $body['headers'] ?? [],
+        'payload' => $body['payload'] ?? '{}'
+    ];
+
+    $response = new Response($res);
+
+    try {
+        $userFunction = include(join_paths($body['path'] ?? DEFAULT_PATH, $body['file'] ?? DEFAULT_FILE));
```
eldadfux, line 71: What is the goal of the `join_paths` function?

runtimes/php-8.0/composer.json.fallback
Meldiron: Our PHP server MUST see `user/function` library. So if user does not have `composer.json`, we use this fallback one.

runtimes/php-8.0/prepare.php
Meldiron: Similar to previous comment. If there is composer.json, it MUST have name `user/function`. So this script makes sure user's composer.json is updated with correct name.

runtimes/php-8.0/server.php
Meldiron: Created PR with fix: https://github.com/open-runtimes/open-runtimes/pull/12
(hard to review, will only be 3 lines change when we merge this PHP runtime. Will fix this in all Node,Deno,PHP)

runtimes/php-8.0/composer.json.fallback
PineappleIOnic: If the user does not add their own composer file we add one for them so when we install dependencies and include the user's code as a package it doesn't crash

runtimes/php-8.0/prepare.php
PineappleIOnic: It will automatically rename the "name" field of the user's composer.json to "user/function" as that is required in the server's composer file to automatically install the user's dependencies. This file makes it so the user themselves does not have to do it.

runtimes/php-8.0/composer.json.fallback
PineappleIOnic: oops, my GitHub didn't refresh with Matej's comment. My bad

PR #14, opened 2022-02-17
Migrate path and file to env variable

Meldiron:


runtimes/deno-1.14/docker-compose.yml
```diff
@@ -10,7 +10,7 @@ services:
     environment:
       - INTERNAL_RUNTIME_KEY
```
christyjacob4, line 11:
```suggestion
      - RUNTIME_KEY
```

In our future iteration let's remember to change these to just RUNTIME_KEY and RUNTIME_ENTRYPOINT

I don't think we're conveying any additional meaning using `INTERNAL` , Lesser words the better 🙂

christyjacob4 (approved):

PR #22, opened 2022-02-18
Updated Travis settings for Deployment

eldadfux:


deploy.sh
```diff
@@ -0,0 +1,2 @@
+echo 'Starting build...'
```
Meldiron, line 1: If we moved deploy.sh, should we also move build.sh?

Meldiron:
Please sync with main, there is now a bunch of new runtimes (versions) in there.

eldadfux:
> Please sync with main, there is now a bunch of new runtimes (versions) in there.

Synced.

christyjacob4 (approved):
Looks good 🤩

PR #26, opened 2022-02-19
Fix: Improved readmes

Meldiron:


runtimes/python-3.8/README.md
```diff
@@ -84,18 +83,17 @@ You can respond with `json()` by providing object:
 
 ```python
 def main(req, res):
-    res.json({
+    return res.json({
```
christyjacob4, line 86: Do we need a return here ?

runtimes/ruby-3.1/README.md
```diff
@@ -18,7 +18,7 @@ printf 'def main(request, response)\n    return response.json({:n => rand()})\ne
 2. Build the code:
 
 ```bash
-docker run --rm --interactive --tty --volume $PWD:/usr/code openruntimes/ruby:3.1 sh /usr/local/src/build.sh
+docker run -e ENTRYPOINT_NAME=index.rb --rm --interactive --tty --volume $PWD:/usr/code openruntimes/ruby:3.1 sh /usr/local/src/build.sh
```
christyjacob4, line 21: Why do we need a `ENTRYPOINT_NAME` 

1. build.sh does not seem to be using this env variable anywhere
2. Didn't we rename it to `INTERNAL_RUNTIME_ENTRYPOINT`  ??

Please check this in the other files as well.

runtimes/ruby-3.0/README.md
```diff
@@ -18,7 +18,7 @@ printf 'def main(request, response)\n    return response.json({:n => rand()})\ne
 2. Build the code:
 
 ```bash
-docker run --rm --interactive --tty --volume $PWD:/usr/code openruntimes/ruby:3.0 sh /usr/local/src/build.sh
+docker run -e ENTRYPOINT_NAME=index.rb --rm --interactive --tty --volume $PWD:/usr/code openruntimes/ruby:3.0 sh /usr/local/src/build.sh
```
christyjacob4, line 21: Why do we need a `ENTRYPOINT_NAME` 

1. build.sh does not seem to be using this env variable anywhere
2. Didn't we rename it to `INTERNAL_RUNTIME_ENTRYPOINT`  ??

Please check this in the other files as well.

runtimes/python-3.10/README.md
```diff
@@ -18,13 +18,13 @@ printf "import random\n\ndef main(req, res):\n    return res.json({'n': random.r
 2. Build the code:
 
 ```bash
-docker run --rm --interactive --tty --volume $PWD:/usr/code open-runtimes/python:3.10 sh /usr/local/src/build.sh
+docker run -e ENTRYPOINT_NAME=main.py --rm --interactive --tty --volume $PWD:/usr/code openruntimes/python:3.10 sh /usr/local/src/build.sh
```
christyjacob4, line 21: Why do we need a `ENTRYPOINT_NAME` 

1. build.sh does not seem to be using this env variable anywhere
2. Didn't we rename it to `INTERNAL_RUNTIME_ENTRYPOINT`  ??

Please check this in the other files as well.

christyjacob4 (changes requested):
Left some questions...

runtimes/python-3.8/README.md
Meldiron: Yes. I was doing QA and function was failing. Doing `return` fixed it. Not sure why.
We do the same in tests and examples, so I think it's safe to also provide it in example.

runtimes/ruby-3.1/README.md
Meldiron: We need `ENTRYPOINT_NAME` for building, and `INTERNAL_RUNTIME_ENTRYPOINT` for running.

It is not needed in Ruby, but it is needed in Rust, Dart or Swift. I decided to put it into all of them for consistency. I think OR should be language agnostic, and I think you should know all runtimes once you learn to use one. Even tho it's not needed for Python or Node, it's good practice to learn to use it, otherwise you will be getting weird errors in build of other runtimes and you would not know what is going on.

christyjacob4: We should keep it only for runtimes that need it. For all other runtimes, it is information that is never used.

runtimes/python-3.8/README.md
christyjacob4: okay 👍

christyjacob4 (approved):

PR #35, opened 2022-02-23
Use `waitress` prod server instead of `flask` development server

abnegate:


runtimes/python-3.10/requirements.txt
```diff
@@ -1 +1,2 @@
-Flask==2.0.1
\ No newline at end of file
+Flask==2.0.1
```
Meldiron, line 1: We still kept Flask dependency

Meldiron (changes requested):
When ready, please apply all changes to all versions of Python runtime.

eldadfux:
Tests are failing

runtimes/python-3.10/requirements.txt
abnegate: Yep Flask is still providing the framework, we are just replacing the web server

eldadfux (approved):

Meldiron (approved):
LGTM

PR #42, opened 2022-02-24
Fix: Improve Swift runtime

Meldiron:


Meldiron (commented):
Looks good, Ill test it manually later today to ensure it works

Meldiron (commented):
Manual QA went well

runtimes/swift-5.5/Sources/App/routes.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 struct RequestValue {
     var path: String = "/usr/code"
-    var file: String = "index.swift"
+    var file: String = "Sources/index.swift"
```
christyjacob4, line 6: Shouldnt this value be set by the environment variable ?  rather than hardcoding it ? 
What if the developer has a different one ?

christyjacob4 (approved):
Looks good 👍

runtimes/swift-5.5/Sources/App/routes.swift
Meldiron: Not sure... Tests use different entrypoint so it works, but I can't see the build process moving anything to the index file. Maybe this variable is not being used.

lohanidamodar: This is definitely just setting the default value and it's fine. we overrite it in line 28 below.

christyjacob4:
@abnegate  Still have some open questions about the Swift entrypoint  

I see we're using different env variables in the .env
https://github.com/open-runtimes/open-runtimes/blob/fix-swift/runtimes/swift-5.5/.env#L2

and the Dockerfile
https://github.com/open-runtimes/open-runtimes/blob/fix-swift/runtimes/swift-5.5/Dockerfile

and the route
https://github.com/open-runtimes/open-runtimes/blob/fix-swift/runtimes/swift-5.5/Sources/App/routes.swift#L6

Shouldn't we be using the value from the environment variables ? Would love your help understanding.

abnegate:
> Shouldn't we be using the value from the environment variables ? Would love your help understanding.

For Swift, the file/path is actually redundant. The values set in `routes.swift` were sane defaults to keep the model consist across languages but are not needed. The only stipulation is that a `main(req: RequestValue, res: RequestResponse)` function is defined in _any_ of the developers files

christyjacob4:
Thank you for the clarification @abnegate . I think we should communicate this in the Appwrite docs for functions.

PR #44, opened 2022-02-25
Fail build if any command fails

abnegate:
- Fixes integration errors where a build step fails but the script exits with a success code


Meldiron (approved):
LGTM

christyjacob4 (approved):

PR #51, opened 2022-03-04
Docs: Node reserved key

Meldiron:
Improved docs of Node runtimes, to mention reserved keyword `code`. People should avoid using this, as this is not consistent behaviour in all runtimes.


TorstenDittmann (approved):

christyjacob4:
@Meldiron  have we made sure to communicate this in the appwrite docs as well ?

PR #54, opened 2022-03-09
Fix: Implemented one-time include

Meldiron:
Problematic code:
```
<?

function myFunc() {
        return "myFunc";
}

return function($req, $res) {
  $res->json([ 'n' => \mt_rand() / \mt_getrandmax() ]);
};
```

Failed due to:
```
Fatal error: Cannot redeclare myFunc() (previously declared in /usr/code-start/index.php:4) in /usr/code-start/index.php on line 3
```

This PR makes sure we only import the code once, and share it between executions. Code above now works.

### DIFF
--- a/runtimes/php-8.0/server.php
+++ b/runtimes/php-8.0/server.php
@@ -68,7 +68,7 @@ $server->on("Request", function($req, $res) use(&$userFunction) {
     $response = new Response($res);
 
     try {
-        if($userFunction == null) {
+        if($userFunction === null) {
             $userFunction = include(USER_CODE_PATH . '/' . getenv('INTERNAL_RUNTIME_ENTRYPOINT'));
         }
 
--- a/runtimes/php-8.0/server.php
+++ b/runtimes/php-8.0/server.php
@@ -30,8 +30,9 @@ class Response {
     }
 }
 
+$userFunction = null;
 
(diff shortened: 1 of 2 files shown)


Meldiron:
Note: `include_once` cannot be used, it would not return the function in second execution.

runtimes/php-8.0/server.php
```diff
@@ -67,7 +68,9 @@ function json($json, $status = 200) {
     $response = new Response($res);
 
     try {
-        $userFunction = include(USER_CODE_PATH . '/' . getenv('INTERNAL_RUNTIME_ENTRYPOINT'));
+        if($userFunction == null) {
```
TorstenDittmann, line 71:
```suggestion
        if($userFunction === null) {
```

runtimes/php-8.1/server.php
```diff
@@ -67,7 +68,9 @@ function json($json, $status = 200) {
     $response = new Response($res);
 
     try {
-        $userFunction = include(USER_CODE_PATH . '/' . getenv('INTERNAL_RUNTIME_ENTRYPOINT'));
+        if($userFunction == null) {
```
TorstenDittmann, line 71:
```suggestion
        if($userFunction === null) {
```

PR #56, opened 2022-03-14
Revert "Docs: Node reserved key"

christyjacob4:
Reverts open-runtimes/open-runtimes#51


Meldiron (approved):
We can revert now, the issue is comming from Appwrite, not Open Runtimes.

PR #57, opened 2022-03-15
Feat: Remove code from error body

Meldiron:
This information is no longer needed in the body. Appwrite uses response code now.


runtimes/deno-1.13/server.ts
```diff
@@ -1,4 +1,4 @@
-import { Application } from "https://deno.land/x/oak@v10.1.0/mod.ts";
+import { Application } from "https://deno.land/x/oak@v8.0.0/mod.ts";
```
christyjacob4, line 1: Why did we decide to downgrade the library. ?

Meldiron: Fixed

christyjacob4 (approved):

PR #58, opened 2022-03-15
Feat: Plain error messages

Meldiron:


Meldiron:
**🚨 This should be merged AFTER feat-deprecate-code-body, and base should be changed to main 🚨**

PR #59, opened 2022-03-15
Feat: Rename http codes

Meldiron:
4XX in Appwrite indicates client error, but this is error in executor communication, happening between servers.


Meldiron:
**🚨 This should be merged AFTER feat-plain-error-messages, and base should be changed to main 🚨**

christyjacob4 (approved):

PR #66, opened 2022-03-25
Feat Kotlin runtime

abnegate:


runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/RuntimeResponse.kt
```diff
@@ -0,0 +1,34 @@
...
+        var mapper: JsonMapper? = null
+
+        @JvmStatic
+        fun error(error: Throwable): RuntimeResponse {
+            return RuntimeResponse("{\"message\": \"" + error.message + "\"}", 500)
+        }
+
+        @JvmStatic
+        fun unauthorized(): RuntimeResponse {
+            return RuntimeResponse("{\"message\": \"Unauthorized\"}", 401)
```
christyjacob4, line 31: https://github.com/open-runtimes/open-runtimes/blob/3261573d3a02e7dbad1868e465d2858a3a183494/runtimes/deno-1.13/server.ts#L13

We should return a 500 status code with a string response body

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/RuntimeResponse.kt
```diff
@@ -0,0 +1,34 @@
...
+        this.statusCode = statusCode
+        return this
+    }
+
+    companion object {
+        var mapper: JsonMapper? = null
+
+        @JvmStatic
+        fun error(error: Throwable): RuntimeResponse {
+            return RuntimeResponse("{\"message\": \"" + error.message + "\"}", 500)
```
christyjacob4, line 26: The response here should be a simple string rather than a JSON string
https://github.com/open-runtimes/open-runtimes/blob/3261573d3a02e7dbad1868e465d2858a3a183494/runtimes/deno-1.13/server.ts#L45

christyjacob4 (changes requested):
Looks good. Just some minor comments

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/RuntimeResponse.kt
```diff
@@ -0,0 +1,34 @@
+package io.openruntimes.kotlin
+
+import io.javalin.plugin.json.JsonMapper
+
+class RuntimeResponse(
+    var data: String = "",
+    var statusCode: Int = 200
+) {
+    fun json(data: Map<String, Any?>, statusCode: Int = 200): RuntimeResponse {
+        this.data = mapper?.toJsonString(data) ?: ""
```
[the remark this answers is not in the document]
abnegate: This is because we can not have the `RuntimeResponse.data` be a `String` _or_ `Map` with static typing, so we opt for the common format and serialise the data to string.

Runtimes that return JSON objects would likely just be serialising in an internal method

abnegate:
@christyjacob4 Updated

christyjacob4:
Amazing work @abnegate  🥳

Issue #63, opened 2022-03-24
Support ES Modules


JakeAi:
Instead of require, use import("path") or require("path").default. This will also allow webpack, typescript, and other bundling options. Typescript was technically supported in the previous functions implementation.

https://github.com/open-runtimes/open-runtimes/blob/main/runtimes/node-17.0/server.js#L25

Meldiron:
My initial idea to support is to do: (pseudo code)
```
if userFunction.default
  response = userFunction.default()
else
  response = userFunction()
```

Ill investigate further.

Meldiron:
A new version has been released and it includes this change 🥳  Can we close this issue? 😇

PR #67, opened 2022-03-25
Feat: ES module support (Node)

Meldiron:
Issue: https://github.com/open-runtimes/open-runtimes/issues/63

If you have `index.ts`:
CleanShot 2022-03-25 at 09 08 27

It will build to `index.js`:

CleanShot 2022-03-25 at 09 08 30

OPR for Node now supports this, and executes properly:

CleanShot 2022-03-25 at 09 08 22

### DIFF
--- a/runtimes/node-14.5/server.js
+++ b/runtimes/node-14.5/server.js
@@ -24,11 +24,15 @@ const server = micro(async (req, res) => {
     try {
         let userFunction = require(USER_CODE_PATH + '/' + process.env.INTERNAL_RUNTIME_ENTRYPOINT);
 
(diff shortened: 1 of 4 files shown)


christyjacob4 (approved):

JakeAi:
If you keep the sanity check, you may want to check if `userFunction.default` is also a function, and I would maybe look into changing the error message to something along the lines of "The exported code must export a function."

christyjacob4:
@JakeAi  I think that's a good suggestion. 
@Meldiron Please incorporate this check as well.

christyjacob4 (changes requested):
@JakeAi  I think that's a good suggestion. 
@Meldiron Please incorporate this check as well.

christyjacob4 (approved):

PR #69, opened 2022-03-28
Feat dotnet runtime

abnegate:


runtimes/dotnet-6.0/README.md
```diff
@@ -0,0 +1,107 @@
...
+public async Task<RuntimeResponse> Main(RuntimeRequest req, RuntimeResponse res) => 
+    res.Json(new() {
+        { "message" , "Hello Open Runtimes 👋" },
+        { "env", req.Env },
+        { "headers", req.Headers },
+        { "payload", req.Payload }
+    });
+```
+
+- To handle dependencies, you need to have `csproj` file containing the `PackageReferences` you desire. Dependencies will be automatically cached and installed, so you don't need to include `node_modules` folder in your function.
```
christyjacob4, line 87: `node_modules` - I believe we need to replace this

christyjacob4 (changes requested):
Some minor changes

PR #70, opened 2022-03-31
ruby: print actual execption on user code failure

lohanidamodar:


Meldiron (approved):
If tests don't cover the topic, it would be great to always include QA screenshots in the description.

Other than that, LGTM.

abnegate (approved):

runtimes/ruby-3.0/server.rb
```diff
@@ -76,7 +76,7 @@ def json(object)
     p e
     status 500
     content_type :json
-    return { code: 500, message: 'File not found or is not a valid ruby file.' }.to_json
+    return { code: 500, message: "Exception Occurred #{e.class}. Message: #{e.message}. Backtrace:  \n #{e.backtrace.join("\n")}" }.to_json
```
christyjacob4, line 79:
```suggestion
    return { code: 500, message: "#{e.backtrace.join("\n")}" }.to_json
```

runtimes/ruby-3.1/server.rb
```diff
@@ -76,7 +76,7 @@ def json(object)
     p e
     status 500
     content_type :json
-    return { code: 500, message: 'File not found or is not a valid ruby file.' }.to_json
+    return { code: 500, message: "Exception Occurred #{e.class}. Message: #{e.message}. Backtrace:  \n #{e.backtrace.join("\n")}" }.to_json
```
christyjacob4, line 79:
```suggestion
    return { code: 500, message: "#{e.backtrace.join("\n")}" }.to_json
```
Just the trace is good enough 👍

christyjacob4 (changes requested):
minor comment

PR #72, opened 2022-04-11
Fix python response code

Meldiron:
- You cannot use status in response: https://github.com/appwrite/appwrite/issues/3064
- You MUST return in python

Both fixed in this PR.

This will now work: `return response.send("Not found", 404)` as well as simply `response.send("Not found", 404)`

### DIFF
--- a/runtimes/python-3.10/server.py
+++ b/runtimes/python-3.10/server.py
@@ -7,10 +7,18 @@ import importlib
 app = Flask(__name__);
 
 class Response:
-    def send(self, text):
-        return text;
-    def json(self, obj):
-        return jsonify(obj);
+    _response = None
+    _status = None
+
+    def __init__(self):
+        self._response = None
+        self._status = None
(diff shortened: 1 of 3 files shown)


runtimes/python-3.10/server.py
```diff
@@ -58,12 +66,14 @@ def handler(u_path):
 
     # Check if function exists
     if userModule is None:
-        return {'message': 'function not found, Did you forget to name it `main`?', 'code': 500}, 500;
+        return {'message': 'function not found, Did you forget to name it `main`?', 'code': 500}, 500
```
christyjacob4, line 69:
```suggestion
        return {'message': 'Function not found. Ensure the function is called `main`?', 'code': 500}, 500
```

christyjacob4 (changes requested):
Please fix the error message in all the python runtimes

christyjacob4 (approved):

PR #74, opened 2022-04-27
Capture console logs from user codes

lohanidamodar:
All runtimes except Swift are working
## Changes
-  Update the model of response from runtimes
- Skipped console logs capturing test in Swift
- Deno only has 1.21 and 1.24 removed other versions
- Removed node 15.5 and 17.0 runtimes
- Removed dart 2.12, 2.13 and 2.14


runtimes/node-17.0/server.js
```diff
@@ -38,8 +43,10 @@ const server = micro(async (req, res) => {
             await userFunction(request, response);
         }
     } catch (e) {
-        send(res, 500, e.code === 'MODULE_NOT_FOUND' ? "Code file not found." : e.stack || e);
+        send(res, 500, {stdout: logs, stderr: e.code === 'MODULE_NOT_FOUND' ? "Code file not found." : e.stack || e});
     }
+    logs.length = 0;
+    console.log = console.stdlog;
 });
```
Meldiron, line 50: This sounds unsafe. Why not `logs = []`, and why are we changing value of `console.log` but not `console.error` and others?

runtimes/node-17.0/server.js
```diff
@@ -17,9 +17,14 @@ const server = micro(async (req, res) => {
         payload: body.payload ?? '',
     };
 
+    console.stdlog = console.log.bind(console);
+    logs = [];
+    console.log = function(){
+        logs.push(Array.from(arguments));
```
Meldiron, line 23: Soo if I low 2 things, are they part of one array, or are they separated?

`console.log(1,3)` is `[1,3]` or [[1,3]]`? Also not sure which one is better.

runtimes/node-17.0/server.js
```diff
@@ -17,9 +17,14 @@ const server = micro(async (req, res) => {
         payload: body.payload ?? '',
     };
 
+    console.stdlog = console.log.bind(console);
```
Meldiron, line 20: can this be some our variable? Sounds scarry to set variable on `console` object

tests/node-17.0/tests.js
```diff
@@ -16,7 +16,7 @@ module.exports = async (req, res) => {
     const payload = JSON.parse(req.payload === '' ? '{}' : req.payload);
 
     const todo = await fetch(`https://jsonplaceholder.typicode.com/todos/${payload.id ?? 1}`).then(r => r.json());
-
+    console.log('stdout', 'other outputs');
```
Meldiron, line 19: In test, can we add multiple logs? also multiple types, maybe even try to log object. For all, lets have test check in PHP

runtimes/node-17.0/server.js
lohanidamodar: if I don't do `Array.from(arguments)` `console.log(1,3)` becomes `{0: 1, 1: 3}`

tests/node-17.0/tests.js
lohanidamodar: haven't gone to the test yet, still trying to get the runtimes right before moving on to test

runtimes/node-17.0/server.js
lohanidamodar: what should I capture, let me know, I just used the StackOverflow reference you provided me and implemented this.

runtimes/node-17.0/server.js
lohanidamodar: it can be, will update to that

runtimes/node-17.0/server.js
Meldiron: CleanShot 2022-04-27 at 09 19 05

There are a few of them: `debug()`, `error()`, `info()`, `log()`, `warn()`. We just need to capture all of them, the same way we do it for `log()`.

runtimes/node-17.0/server.js
Meldiron: Let's keep it for internal use

Meldiron:
- If you log array/object, it does not show actual value
- Let's join with ` ` arguments, and then logs with `\n`

lohanidamodar:
@Meldiron need another set of reviews

test.http
```diff
@@ -0,0 +1,7 @@
+POST http://0.0.0.0:3000
+content-type: application/json
+X-internal-Challenge: test-secret-key
+
+{
+    "garbage": "garbage"
```
Meldiron, line 6: Maaaybe a different word here 😅

tests/dart-2.16/lib/tests.dart
```diff
@@ -15,17 +15,17 @@ import 'package:dio/dio.dart' hide Response;
...
 Future<void> start(final req, final res) async {
-
   final payload = jsonDecode(req.payload == '' ? '{}' : req.payload);
 
   final id = payload['id'] ?? '1';
-  final todo = await Dio().get('https://jsonplaceholder.typicode.com/todos/$id');
-
+  final todo =
+      await Dio().get('https://jsonplaceholder.typicode.com/todos/$id');
+  print('stdout');
```
Meldiron, line 23: Can we add more prints to test? primitive types + object + array. If behavior is different from other runtimes, it might be fine, we will do a specific test for that runtime. The main idea is to fully understand what will be returned.

test.http
lohanidamodar: this is something to be deleted, i made this file for local testing. I took this from test files

lohanidamodar:
@abnegate Swift still seems failing please check!

runtimes/dart-2.12/server.dart
```diff
@@ -25,12 +32,34 @@ void main() async {
       );
 
       final response = Response();
-      await user_code.start(request, response);
-      return shelf.Response.ok(response.body);
+      await runZonedGuarded(
+        () async {
+          await user_code.start(request, response);
+        },
+        (e, stackTrace) => print('Oh noes! $e $stackTrace'),
```
Meldiron, line 39: Oh noes on purpose?

Meldiron: It's in like 5 places.

lohanidamodar: copy paste 😄

lohanidamodar: fixed

Meldiron (approved):
Looks great

.travis.yml
```diff
@@ -53,24 +59,24 @@ env:
...
+    #   ENTRYPOINT=tests.ts
+    #   SERVER_PROCESS=server.ts
+    #   IMAGE=openruntimes/deno:${VERSION}-1.13
+    #   ARCH=linux/amd64
+    # - RUNTIME=deno-1.12
+    #   PHP_CLASS=Deno112
+    #   ENTRYPOINT=tests.ts
+    #   SERVER_PROCESS=server.ts
+    #   IMAGE=openruntimes/deno:${VERSION}-1.12
+    #   ARCH=linux/amd64
```
christyjacob4, line 79: I believe this needs to be uncommented ?

runtimes/deno-1.21/server.ts
```diff
@@ -21,14 +21,33 @@ app.use(async (ctx) => {
...
+        }
+    });
+    logs.push(args.join(" "));
+  }
+
   const response = {
     send: (text: string, status = 200) => {
       ctx.response.status = status;
-      ctx.response.body = text;
+      ctx.response.body = {response: text, stdout: logs.join('\n')};
```
christyjacob4, line 46: In the other runtimes ( c++, dart ) , we are sending the response body as a json encoded string. Here we're directly sending a JSON. I don't think this is expected.

runtimes/dotnet-3.1/src/RuntimeResponse.cs
```diff
@@ -28,7 +27,7 @@ namespace DotNetRuntime
             Dictionary<string, object?> data,
             int statusCode = 200)
         {
-            Data = JsonSerializer.Serialize(data);
+            Data = data;
```
christyjacob4, line 30: I'm seeing different patterns being followed for different runtimes.
In some cases we're sending serialised strings, and in some we're sending the JSON object as it is. 

Please test all the runtimes if this is working

runtimes/deno-1.21/server.ts
christyjacob4: Please test all the runtimes to **verify** this is working

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
```diff
@@ -28,9 +30,34 @@ suspend fun execute(ctx: Context) {
...
+    System.setErr(userErr)
+
     try {
-        ctx.result(codeWrapper.main(request, response).data)
+        val userResponse = codeWrapper.main(request, response)
+        val output = mutableMapOf(
+            "response" to userResponse.data,
+            "stdout" to outStream.toString()
+        )
+        ctx.json(output)
```
christyjacob4, line 49: Doesn't the output need to be serialised?  We are doing it in Java, and the others

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
```diff
@@ -28,9 +30,34 @@ suspend fun execute(ctx: Context) {
...
     } catch (e: Exception) {
-        ctx.status(500).result(e.stackTraceToString())
+        e.printStackTrace()
+        val output = mutableMapOf(
+            "stdout" to outStream.toString(),
+            "stderr" to errStream.toString()
+        )
+        ctx.status(500).json(output)
```
christyjacob4, line 56: similarly here as well

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
christyjacob4: @abnegate  helped resolve the comment 👍

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
christyjacob4: @abnegate  helped resolve the comment 👍

runtimes/swift-5.5/Sources/App/routes.swift
```diff
@@ -110,19 +100,51 @@ func routes(_ app: Application) throws {
...
         }
 
+        setvbuf(stdout, nil, _IONBF, 0)
+        setvbuf(stderr, nil, _IONBF, 0)
+
+        let outPipe = Pipe()
+        let errPipe = Pipe()
+        var outString = ""
+        var errString = ""
+
```
christyjacob4, line 110: Since console logs are not working in swift, can we revert these changes in swift to simply return an empty string for `stdout`?

christyjacob4 (commented):
PR looks great. Please validate the comments and uncomment the tests for deno and we're ready to go 👍

runtimes/deno-1.21/server.ts
lohanidamodar: I think we should add new runtimes in deno and discard the old ones. I'll look at the releases, the old ones I don't think it's worth spending much time on them, so I've commented those out.

runtimes/swift-5.5/Sources/App/routes.swift
lohanidamodar: Well I'm okay either way. as it's not working

.travis.yml
lohanidamodar: I've rather added new Deno 1.24 runtime, don't think maintaining and updating all these old deno runtimes is fruitful. Deno has lot newer releases `deno 1.21` and `deno 1.24`. let me know what you think

runtimes/deno-1.21/server.ts
christyjacob4: In that case, let's remove the older ones and finalise this PR

.travis.yml
christyjacob4: yes, this is a tedious process maintaining a lot of similar releases and is not worth it... 
Let's remove the old ones and keep just the latest 1 and an LTS

runtimes/swift-5.5/Sources/App/routes.swift
christyjacob4: @abnegate please address this when you get the time.

tests/Deno124.php
```diff
@@ -6,6 +6,6 @@
 // PHP class: Deno114
```
lohanidamodar, line 6:
```suggestion
// PHP class: Deno124
```

christyjacob4 (approved):

christyjacob4:
Please create a corresponding PR on appwrite/runtimes to fetch the new versions

PR #84, opened 2022-07-13
Feat: Improve caching in Deno 1.21

Meldiron:
`deno cache` didn't really serve a purpose here. We have seen extremely long cold starts, most likely due to the re-downloading of dependencies on the first request:

image

This new `vendor` approach is downloading dependencies into `vendor` folder, making sure all is downloaded during the build step.

---

This PR also moves "check" into the build step, and runs execution with `--no-check` to improve cold start speed even more.

### DIFF
--- a/runtimes/deno-1.21/build.sh
+++ b/runtimes/deno-1.21/build.sh
@@ -6,19 +6,14 @@ set -e
 # Prepare separate directory to prevent changing user's files
 cp -R /usr/code/* /usr/builds
 
-# Prepare Deno cache folder
-mkdir -p /usr/builds/deno-cache
 
-# Set Deno Cache directory
-export DENO_DIR="/usr/builds/deno-cache"
-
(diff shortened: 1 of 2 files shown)


TorstenDittmann (commented):
Will this also include user dependencies that are not part of an import_map?

Meldiron:
All user dependencies are fetched into vendor (and import_map) during the build process from exactly the same source code that later runs during execution. We should have all dependencies.

And no, they won't be fetched later. It will throw an error about not being able to find a library file.

Meldiron:
As a test I tried to make my entrypoint `mod.ts` that uses method from `newfile.ts` that imports a library. During build, this library was also stored. As long as your code uses a library, it should always be downloaded to vendor.

CleanShot 2022-07-22 at 09 10 58@2x

loks0n:

Issue #86, opened 2022-07-21
fix: Python packages fail to install


gijsvanlokven:
Some Python packages cannot be installed due to build dependencies that are missing in the Alpine image.
Most notable being Numpy, and without Numpy packages like Pandas, Pytorch, Scikit-learn, Tensorflow will not install.

While it is an option to switch to the Ubuntu Python image, adding the needed dependencies to Alpine is still more space efficient.

Currently working on a fix by adding needed build dependencies to the Dockerfiles of the Python runtimes

No action needed, just opening an issue for the pull request

malachibazar:
Is there any news on this? I'm trying to use OpenAI's module, which requires Numpy.

gijsvanlokven:
@malachibazar The fix has been accepted so it'll probably come out with the next merge.
If you need the fix with any speed just add RUN apk --no-cache add musl-dev linux-headers g++ to the python docker files right under the FROM line

loks0n:
This was filed against v2-era runtimes whose versions have since been removed; the runtimes have been rewritten under the current versions/latest layout. Closing as outdated — please open a fresh issue if you can reproduce on current images.

PR #87, opened 2022-07-21
Fix 86 python package dependencies

gijsvanlokven:
Pip could not install Numpy on the Alpine image because of missing build dependencies.
Added the needed dependencies to the Dockerfiles of all Python Runtimes.


gijsvanlokven:
@eldadfux @Meldiron would either of you mind checking this over? Been running these versions of the python runtimes without issue since the request was opened!

abnegate (approved):
Looks good!

gijsvanlokven:
Fix for issue #86

stnguyen90:
FYI, I tested this with python 3.10 and it increased the image size from 51MB to 243MB 👀

PR #88, opened 2022-08-22
Fix: Null vars

Meldiron:
In Swift, we noticed all variables are missing if there is at least one that has a value of `null`. This PR addresses the problem and implements tests to ensure all other runtimes work properly.

Related issues:

- https://github.com/appwrite/appwrite/issues/3707

### DIFF
--- a/docs/add-runtime.md
+++ b/docs/add-runtime.md
@@ -221,6 +221,7 @@ Within the folder you will need to create a function for your runtime that will 
     "message": "Hello Open Runtimes 👋",
     "header": req.headers['x-test-header'],
     "env": req.env['test-env'],
+    "null-env": req.env['null-env'],
     "todo": {{body from your todo API http response}},
 ```
 
--- a/runtimes/java-11.0/src/main/java/io/openruntimes/java/RuntimeResponse.java
+++ b/runtimes/java-11.0/src/main/java/io/openruntimes/java/RuntimeResponse.java
@@ -1,6 +1,7 @@
 package io.openruntimes.java;
 
 import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
 
 import java.util.Map;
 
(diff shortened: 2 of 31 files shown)


Meldiron:
Works:

```
curl -H "X-Internal-Challenge: secret-key" -H "Content-Type: application/json" -X POST http://localhost:3000/ -d '{"env":{"SECRET_KEY":"klucik"},"payload":"Body","timeout":15}'
```

---

Does not work:

```
curl -H "X-Internal-Challenge: secret-key" -H "Content-Type: application/json" -X POST http://localhost:3000/ -d '{"env":{"SECRET_KEY":"klucik","APPWRITE_FUNCTION_EVENT":null},"payload":"Body","timeout":15}'
```

---

With code in `runtimes/swift-5.5/example/Sources/index.swift`:

```swift
import Foundation
import AsyncHTTPClient

func main(req: RequestValue, res: RequestResponse) async throws -> RequestResponse {
    var todoId: String = "1"

    if !req.payload.isEmpty,
        let data = req.payload.data(using: .utf8),
        let payload = try? JSONSerialization.jsonObject(with: data, options: []) as? [String: Any] {
            todoId = payload["id"] as? String ?? "1"
    }

    let httpClient = HTTPClient(eventLoopGroupProvider: .createNew)
    let request = HTTPClientRequest(url: "https://jsonplaceholder.typicode.com/todos/\(todoId)")
    let response = try await httpClient.execute(request, timeout: .seconds(30))
    let data = try await response.body.collect(upTo: 1024 * 1024)
    let todo = try JSONSerialization.jsonObject(with: data, options: []) as? [String: Any]

    let secretKey = req.env["SECRET_KEY"] ?? "SECRET_KEY environment variable not found. You can set it in Function settings."

    return res.json(data: [
        "message": "Hello Open Runtimes 👋",
        "todo": todo,
        "secretKey": secretKey,
        "env": req.env
    ])
}
```

> Notice `secretKey` and `env` in the response.

wess (approved):
Nice update on Swift, but also love the other language and test updates.

Meldiron:
Resolved by v3, and we also now have test for missing variable

PR #91, opened 2022-09-12
Rename all 'env'to 'variables'

Meldiron:
⚠️ This change requires OPR v3

Since OPR came into existance (Appwrite 0.13+), variables should be read from request object, NOT from environment variables.

This PR makes that obvious by renaming `req.env` to `req.variables` in all runtimes.

- [x] Tests updated

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,8 +139,8 @@ All requests should also have JSON body with the following structure:
 ```json5
 {
     // Following will be exposed to the function
-    "env": {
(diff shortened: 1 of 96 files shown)


gewenyu99:
YES!!!!!!!!!! 5🌟s for Matej. You get cookies

eldadfux:
We should ask GitHub to add a cookie emoji.

PR #102, opened 2022-10-09
Feat: Add basic support for Kubernetes

gijsvanlokven:
- Each runtime has 2 new files added to its directory, a deployment and a service. These can be run using the kubectl -f deployment/service.yaml command to bring up the deployment and accompanying service. kubectl port-forward can then be used to map ports to the host machine.

### DIFF
--- a/runtimes/cpp-17/deployment.yaml
+++ b/runtimes/cpp-17/deployment.yaml
@@ -1,0 +1,38 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: open-runtimes-cpp-17
+  labels:
+    app.kubernetes.io/name: open-runtimes-cpp-17
+    app.kubernetes.io/component: backend
+spec:
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: open-runtimes-cpp-17
+      app.kubernetes.io/component: backend
+  replicas: 1
+  template:
+    metadata:
+      labels:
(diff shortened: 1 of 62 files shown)


stnguyen90:
@gijsvanlokven, thanks for creating this PR! 🙏🏼 I don't think we want to move forward with this right now because it would add a lot more for us to maintain. Perhaps this is better in a separate community repo.

Meldiron:
Thank you so much for the PR 🤩. We're adding the `hacktoberfest-accepted` label to ensure this PR counts towards your Hacktoberfest contributions count. With that said, please stay active on this PR to address any comments once you receive a review. Happy Hacktoberfest! 🎃

gijsvanlokven:
@stnguyen90 in what way would you reckon this would be a lot to maintain? Each runtime just needs 2 extra files. If it's just about the extra files it wouldn't be much trouble to automate the generation of the Kubernetes config files. Would love to hear your thoughts

stnguyen90:
> in what way would you reckon this would be a lot to maintain?

We would need to make sure the files work going forward.

Another thing is K8s is more for orchestration so it wouldn't really fit into this repo. We suggest having a community repo for these files rather than adding these into this repo. I can leave the `hacktoberfest-accepted`, but I don't think we'll be merging this into the open runtimes repo.

stnguyen90:
Closing as we're not looking to include this in this repo.

PR #112, opened 2022-10-28
New architecture diagram

eldadfux:


christyjacob4 (approved):
Looks good. Only improvement I would suggest is the proxy being connected directly to an `executor` rather than a `host`

Meldiron (approved):
Looks good. I agree with Christy's comment

abnegate (approved):

PR #116, opened 2022-11-01
feat: adds open-runtimes example for `cpp 20` stable release

afzal442:
This PR adds open-runtimes example for stable release of `cpp 20`


stnguyen90:
FYI, we're having discussions internally about how to decide which runtime versions we want to support. If we do decide to support cpp 20, this will need to include tests to make sure the CI job passes before it can be merged.

afzal442:
Thanks @stnguyen90 😃 Looking forward to adding CI tests for this soon.

abnegate (changes requested):
Hey @afzal442 looks like the this runtime is still detecting a max of C++ 2017 support:

```
-- The CXX compiler identification is GNU 11.2.1
...
-- Looking for C++ include coroutine
-- Looking for C++ include coroutine - not found
...
-- use c++17
```

You can check this by running the tests and checking the output:

```sh
export RUNTIME=cpp-20; export PHP_CLASS=Cpp20; export ENTRYPOINT=tests.cc; sh tests.sh
```

When C++20 is supported the message will be `-- use c++20`

afzal442:
Thanks @abnegate can u plz suggest what I missing here? Overhauling. ☂️ 😉

abnegate:
@afzal442 I can't say exactly, but the issue is with the detection of the `coroutine` header from the given GCC version (11.2.1)

Meldiron:
Hey there 👋 There were many changes due to v3 of Open Runtimes. We will be adding C++ 20 as part of this PR: https://github.com/open-runtimes/open-runtimes/pull/178

Closing this PR, as it would be very-very hard to apply all new changes.

PR #126, opened 2022-12-19
Feat: Implement no response test

Meldiron:
Create consistent responses for executions with no user-made response.
Also implement proper tests.


Meldiron:
Marking as draft as it might not be relevant, we will have a disucssion soon.

Meldiron:
Not relevant anymore. Addressed in: https://github.com/open-runtimes/open-runtimes/pull/130

PR #129, opened 2023-01-09
Fix: Dart error handling

Meldiron:
Runtime error in user code in Dart caused timeout error. This PR fixes this. It also fixes problem in Appwrite when logs werent visible after error in Dart (related to timeout issue)


lohanidamodar (approved):

Meldiron:
https://github.com/open-runtimes/open-runtimes/pull/177

PR #130, opened 2023-01-10
Feat: Node v3

Meldiron:
Node implementation of V3.
Parent PR: https://github.com/open-runtimes/open-runtimes/pull/136


tests/Node180.php
```diff
@@ -6,6 +6,8 @@
 // PHP class: Node180
 // Entrypoint: tests.js
 
-class Node180 extends Base
+// export RUNTIME="node-18.0"; export PHP_CLASS="Node180"; export ENTRYPOINT="tests.js"; sh tests.sh
```
Meldiron, line 9: leftover

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,57 @@
...
 
-module.exports = async (req, res) => {
-    const payload = JSON.parse(req.payload === '' ? '{}' : req.payload);
+    switch (action) {
+        case 'plaintextResponse':
+            return context.res.send('Hello World 👋');
+        case 'jsonResponse':
+            return context.res.json({ json: true, message: 'Developers are awesome.' });
+        case 'fileResponse':
+            return context.res.file(__dirname + '/resources/picture.png', undefined, { 'content-type': 'image/png' });
```
Meldiron, line 12: Discuss. Ideally no need for `__dirname`, ideally uses `_resources` and not the copy

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,125 @@
...
+            headers: headers,
+            method: req.method,
+            url: req.url
+        },
+        res: {
+            headers: {},
+            body: '',
+            status: 200,
+
+            send: function (body, status, headers) {
```
eldadfux, line 35: I wonder if we want to add more access methods to make the interface even nicer, maybe take some inspiration from Utopia.

Meldiron: Compared to utipia, we are missing: html(), jsonp() anf iframe()

From those, I would agree we should add `html`. Tho I would prefer to avoid others until community requests them. I would prefer to keep it simple. In the end, it's all just alias for `send()`.
WDYT? @eldadfux

eldadfux: I was more in the direction of getHeader, addHeader and stuff like that

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,125 @@
...
+const { text, json, send } = require("micro");
 
 const USER_CODE_PATH = '/usr/code-start';
 
 const server = micro(async (req, res) => {
-    const body = await json(req);
-
-    if (req.headers[`x-internal-challenge`] !== process.env['INTERNAL_RUNTIME_KEY']) {
-        return send(res, 500, {stderr: 'Unauthorized'});
+    if (req.headers[`x-openruntimes-secret`] !== process.env['INTERNAL_RUNTIME_KEY']) {
```
Meldiron, line 8: x-open-runtime

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,71 @@
...
+        case 'redirectResponse':
+            return context.res.redirect('https://github.com/');
+        case 'emptyResponse':
+            return context.res.empty();
+        case 'noResponse':
+            context.res.send('This should be ignored, as it is not returned.');
+            break;
```
abnegate, line 19: This case will not be possible for languages if a return type is defined

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,71 @@
...
+        case 'statusResponse':
+            return context.res.send('FAIL', 404);
+        case 'complexResponse':
+            context.res.setCode(201);
+            context.res.setHeaders({ 'header1': 'value1' })
+            context.res.addHeader('header2', 'value2');
+            context.res.setBody('👌');
+            context.res.setContentType('application/json');
+            context.res.addCookie('cookie1', 'value1');
+            context.res.addCookie('cookie2', 'value2', 'Wed, 21 Oct 2015 07:28:00 GMT', 2592000, '/', 'google.com', true, true, 'Lax');
```
abnegate, line 34: With this approach, `context.res` serves two separate purposes, I think there could be two separate types: one that lets you send a response (has `send`, `json`, `redirect`, etc.) and one that lets you build a response _to_ send (has `setStatusCode`, `setHeaders`, `setBody`, etc. )

abnegate (commented):
Thoughts on dropping `res` and `req` from `context`? For example:

```js
context.send('OK');
```
```js
// Get request header
context.header('x-req-header');

// Set response header
context.header('x-res-header', 'value')

context.json({ 
  'requestBody': context.body(),
  'requestUrl': context.url(),
  'requestMethod': context.method()
});
```

tests/node-18.0/tests.js
Meldiron: Good point. I would still like to have this test otherwise there would be un-tested edge case. In those languages, I simply return 204 and write a commend to explain it.

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,226 @@
...
+        contentType: null,
     };
 
-    console.stdlog = console.log.bind(console);
-    console.stderror = console.error.bind(console);
-    console.stdinfo = console.info.bind(console);
-    console.stddebug = console.debug.bind(console);
-    console.stdwarn = console.warn.bind(console);
+    const contentType = req.headers['content-type'] ?? 'text/plain';
+    const body = contentType === 'application/json' ? await json(req) : await text(req);
```
christyjacob4, line 24: What about other content types? What if a user wants to handle a form submission with an Appwrite function? In this case the content type would be either `multipart/form-data` or `application/x-www-form-urlencoded`

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,226 @@
...
 
-    const response = {
-        send: (text, status = 200) => send(res, status, {response: text, stdout: logs.join('\n'), stderr: errors.join('\n')}),
-        json: (json, status = 200) => send(res, status, {response: json, stdout: logs.join('\n'), stderr: errors.join('\n')}),
+            getContentType: function() {
+                return response.contentType
+            },
+
+            getCookies: function() {
+                return response.cookies;
```
christyjacob4, line 52: shouldn't this be `req.cookies` ?

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,226 @@
...
-    }
+            getStatusCode: function() {
+                return response.statusCode;
+            },
 
-    const response = {
-        send: (text, status = 200) => send(res, status, {response: text, stdout: logs.join('\n'), stderr: errors.join('\n')}),
-        json: (json, status = 200) => send(res, status, {response: json, stdout: logs.join('\n'), stderr: errors.join('\n')}),
+            getContentType: function() {
+                return response.contentType
```
christyjacob4, line 48: IIRC there can be a `content-type` header in both a request and response.

In a request, the `content-type` header tells the the server the type of the request.
In a response, the `content-type` headers tells the browser or resulting client the type of the response. 

Don't we need to have getters for both the content types ?

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,71 @@
...
-        'json(obj, status)' - function to return JSON response. Status code defaults to 200
-    
-    If an error is thrown, a response with code 500 will be returned.
-*/
+module.exports = async (context) => {
+    const action = context.req.headers['x-action'];
 
-module.exports = async (req, res) => {
-    const payload = JSON.parse(req.payload === '' ? '{}' : req.payload);
+    switch (action) {
```
christyjacob4, line 6: We should also be able to support the `text/html` content type. And we need some tests for that

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,71 @@
...
+            context.res.setStatusCode(201);
+            context.res.setHeaders({ 'header1': 'value1' })
+            context.res.addHeader('header2', 'value2');
+            context.res.setBody('👌');
+            context.res.setContentType('application/json');
+            context.res.addCookie('cookie1', 'value1');
+            context.res.addCookie('cookie2', 'value2', 'Wed, 21 Oct 2015 07:28:00 GMT', 2592000, '/', 'google.com', true, true, 'Lax');
 
-    const todo = await fetch(`https://jsonplaceholder.typicode.com/todos/${payload.id ?? 1}`).then(r => r.json());
+            context.res.setBody(JSON.stringify({
```
christyjacob4, line 36: The setBody on L36 overrides the one on L31 so L31 is essentially not being used. Is this expected?

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,71 @@
...
+        case 'requestMethod':
+            return context.res.send(context.req.method);
+        case 'requestUrl':
+            return context.res.send(context.req.url);
+        case 'requestHeaders':
+            return context.res.json(context.req.headers);
+        case 'requestBodyPlaintext':
+            return context.res.send(context.req.body);
+        case 'requestBodyJson':
+            return context.res.send(context.req.body.data ?? 'Missing key');
```
christyjacob4, line 55: we also need tests for cases where the request is of type `multipart/form-data` or `application/x-www-form-urlencoded`

tests/node-18.0/tests.js
Meldiron: Expected. A `getBody` is done inside. That validates (as part of test) both set and get body works.

Meldiron:
> Thoughts on dropping `res` and `req` from `context`?

After internal discussion we decided to keep the separation. Reasons:
- Trying to stay consistent with existing protocols which people already know. Separation into a request and response gives clear indication of what data to expect there
- Trying to avoid magic, at all cost. One method that can do 2 actions based on amount of parameters might be confusing. Also allows easy bugs when `set` and `get` is not clearly written in the code - imagine reviewing such a code. You would need to focus on amount of parameters to decide if logic is correct

tests/node-18.0/tests.js
Meldiron: Added support for `context.res.html('<h1>Your HTML code</h1>)`

tests/node-18.0/tests.js
Meldiron: Added support for form-data and url-encoded body.

runtimes/node-18.0/server.js
Meldiron: Added support for form-data and url-encoded body.

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,239 @@
...
+                        }
+                        if (cookie.secure) {
+                            attributes.push(`Secure`);
+                        }
+                        if (cookie.httponly) {
+                            attributes.push(`HttpOnly`);
+                        }
+
+                        return attributes.join('; ');
+                    }).join('; ');
```
Meldiron, line 144: Is this nessessary for first iteration? Devs can still set cookies on their own with headers. This is making it simpler, but more complex on the Open Runtimes side. I would prefer having this as improvement, alongside better body parsing (XML,formData,..)

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,138 @@
...
+        output = context.res.send('', 500);
+    }
+
+    if(output === null || output === undefined) {
+        context.error('Return statement missing. return context.res.empty() if no response is expected.');
+        output = context.res.send('', 500);
+    }
+
+    output.body = output.body ?? '';
+    output.statusCode = output.statusCode ?? 204;
```
christyjacob4, line 121: The default status code should be 200

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,138 @@
...
+                if (statusCode !== undefined) { response.statusCode = statusCode; }
+                if (headers !== undefined) { response.headers = headers; }
+
+                return {
+                    body: response.body,
+                    statusCode: response.statusCode,
+                    headers: response.headers
+                }
+            },
+            json: function (obj, statusCode, headers = {}) {
```
christyjacob4, line 53: We need to set a default status code for json to be 200

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,138 @@
...
+        res.setHeader(header, output.headers[header]);
     }
 
-    console.log = console.stdlog;
-    console.error = console.stderror;
-    console.debug = console.stddebug;
-    console.warn = console.stdwarn;
-    console.info = console.stdinfo;
+    res.setHeader('x-open-runtimes-logs', encodeURIComponent(logs.join('\n')));
+    res.setHeader('x-open-runtimes-errors', encodeURIComponent(errors.join('\n')));
```
christyjacob4, line 133: We wanted to support logs streaming at one point. Im not sure headers would be the right way to send logs in that case. Its something to think about

runtimes/node-18.0/server.js
```diff
@@ -1,85 +1,138 @@
...
+            output = await userFunction(context);
         }
     } catch (e) {
-        send(res, 500, {stdout: logs.join('\n'), stderr: errors.join('\n') + "\n" + e.code === 'MODULE_NOT_FOUND' ? "Code file not found." : e.stack || e});
+        context.error(e.code === 'MODULE_NOT_FOUND' ? "Code file not found." : e.stack || e);
+        output = context.res.send('', 500);
+    }
+
+    if(output === null || output === undefined) {
+        context.error('Return statement missing. return context.res.empty() if no response is expected.');
```
christyjacob4, line 116: Weren't we going to throw an exception ?

runtimes/node-18.0/server.js
Meldiron: Only possibility in HTTP that allows streaming (as far as I am aware) is body. We surely dont want to touch body and define open-runtimes-specific structure.

I believe streaming support can be added in future with a reserved endpoint.

runtimes/node-18.0/server.js
Meldiron: `context.error` prints a log into errors, so you know what you did wrong.
Line just below this sends 500 response with empty body (500 will mark execution as failed in Appwrite)

runtimes/node-18.0/server.js
Meldiron: Good catch. This was 204 for when you forget to return. Such behaviour no longer exists as not returning throws exception. Updated.

runtimes/node-18.0/server.js
Meldiron: Added proper default values to all methods under `res`

tests/BaseV3.php
```diff
@@ -0,0 +1,264 @@
...
+    {
+        $response = $this->execute(headers: ['x-action' => 'plaintextResponse']);
+        self::assertEquals(200, $response['code']);
+        self::assertEquals('Hello World 👋', $response['body']);
+    }
+
+    public function testJsonResponse(): void
+    {
+        $response = $this->execute(headers: ['x-action' => 'jsonResponse']);
+        self::assertEquals(200, $response['code']);
```
christyjacob4, line 83: Hmm how is this test passing ? Looking at the code, a 204 response is returned

tests/BaseV3.php
```diff
@@ -0,0 +1,264 @@
...
+    }
+
+    public function testHeadersResponse(): void
+    {
+        $response = $this->execute(headers: ['x-action' => 'headersResponse', 'x-open-runtimes-custom-in-header' => 'notMissing']);
+        self::assertEquals(200, $response['code']);
+        self::assertEquals('OK', $response['body']);
+        self::assertEquals('first-value', $response['headers']['first-header']);
+        self::assertEquals('missing', $response['headers']['second-header']);
+        self::assertArrayNotHasKey('x-open-runtimes-custom-out-header', $response['headers']);
```
christyjacob4, line 130: So custom headers cant be sent as response?

tests/BaseV3.php
Meldiron: Ignore 204 you noticed earlier. It was left-over code from earlier iteration that never occured. Such scehario was for not returning at all, which now results in 500 response tested here: https://github.com/open-runtimes/open-runtimes/blob/d7549b672133696085ac26e4c3da71877cecca12/tests/BaseV3.php#L108-L114

tests/BaseV3.php
Meldiron: They can, but they cannot start with `x-open-runtimes-`. Notice `first-header` has test for a proper value from: https://github.com/open-runtimes/open-runtimes/blob/d7549b672133696085ac26e4c3da71877cecca12/tests/node-18.0/tests.js#L22

Meldiron: `x-open-runtimes` is reserved keyword here. Same happens when request comes in. `context.req.headers` will not include any of headers starting with `x-open-runtimes-`.

Code for filtering request headers:

https://github.com/open-runtimes/open-runtimes/blob/3471b94b41f881df4f12cb1745afc9d63ebdfe94/runtimes/node-18.0/server.js#L28-L31

Code for filtering response headers:

https://github.com/open-runtimes/open-runtimes/blob/3471b94b41f881df4f12cb1745afc9d63ebdfe94/runtimes/node-18.0/server.js#L124-L130

tests/node-18.0/tests.js
```diff
@@ -1,37 +1,55 @@
 const fetch = require("node-fetch");
```
christyjacob4, line 2: This file makes more sense under `resources/functions`

Also let's rename `_resources` to `resources`

Meldiron: Resources in this case is supposed to be in context of "resources for functions", not "resources for tests".

To make this clearer, we could have the `picture.png` in `tests/node-18.0/picture.png`, but that would make a lot of duplication of big files (repo will download for long time). That's why I prefered such a shared `resources` folder.

It was named `_resources` due to problems with gitignore file. Ill investigate that.

Meldiron: Folder renamed from `_resources` to `resources`

tests/BaseV3.php
Meldiron: Developers can sand almost any headers. You can do `a`, `bla`, `what-ever-you-want`.

What you cannot do is header like `x-open-runtimes-something-bla`. That will not be passed into context for security reasons.

Regarding `x-open-runtimes-secret`, yes, that needs to be provided otherwise 401 is shown. Tho again, such a secret should bot be visible in the context. So value of this header is removed from context in order for developer to not see this secret.
For example, in Appwrite secret is between runtime and executor. There is no need for developer to know it. Even worse, if developer knew the secret, it would be security vulnerability possibly.

runtimes/node-18.0/server.js
christyjacob4: Okay. Problem for later 👍

runtimes/node-18.0/server.js
christyjacob4: Okay 👍

tests/node-18.0/tests.js
christyjacob4: Well this is very similar to the case in Appwrite. We have tests ( which are the PHP files ) and tests have resources ( which can be functions, images etc etc. ) 

Here's what I'm suggesting

```
tests/
    - BaseV3.php
    - Node180.php
    - resources/
        - images/
            - picture.png
        - functions/
            - node18.js
            - node16.js     
```

Meldiron: Updated. Actually, we no longer need `images` folder as this test and response was removed.

christyjacob4 (approved):

PR #131, opened 2023-01-13
Feat: custom build command

Meldiron:
## Example Usage

`sh -c "sh /usr/local/src/build.sh 'MY_INSTALL_COMMAND' 'MY_BUILD_COMMAND'"` (executor will be adding those parameter to the command)

## Description

Allowing custom build commands adds many possobilities:

### 1. Private dependencies:

CleanShot 2023-01-13 at 13 20 17

### 2. Typescript (and others)

CleanShot 2023-01-13 at 13 04 42

Notes:

- Separate commands for build and install to make it clear what a custom command would be overwriting. Imagine if there was a "build command" which defaulted to "npm install". If you wanted to do custom command, you would need to do something like "tsc && npm install". Devs would not know they are getting rid of "npm install". And we cant expect every function to need npm install, so that has to be overridable also. The best conclusion I was able to find was to allow customization of install and build command separatelly.

## Implementation:

- [x] Node
- [x] PHP
- [x] Ruby
- [x] Deno
- [x] Dart
- [x] C++
- [x] Java
- [x] Python
- [x] Swift
- [x] Dotnet
- [x] Kotlin

## Tests:

- [ ] Install command test
- [ ] Build command test
- [ ] Fallback file test


christyjacob4 (commented):
Approach looks good. Let's proceed.

runtimes/node-18.0/build.sh
```diff
@@ -8,13 +8,21 @@ cp -R /usr/code/* /usr/builds
...
+# Ensure package.json exists
+if [[ ! -f "package.json" ]]; then
+    mv /usr/local/src/package.json.fallback /usr/builds/package.json
 fi
 
-mkdir -p node_modules
+INSTALL_COMMAND=${1:-'npm install'}
+BUILD_COMMAND=${2:-''}    
+
+eval "$INSTALL_COMMAND"
```
eldadfux, line 20: I think we're missing the goal here. I think Open Runtimes should only accept and entrypoint. In the entrypoint we can pass a chain of commands for either build or install from an external platform (e.g. Appwrite). This will allow runtimes to behave exactly the same for both build and execution.

In that external platform (like Appwrite) we can perform validation and ensure that the final command for execution is starting an HTTP server or in the build that the first command is installation or whatever we need.

This PR should actually remove a lot of code rather than adding.

eldadfux (changes requested):

Meldiron:
Closing in favour of: https://github.com/open-runtimes/open-runtimes/pull/154

PR #134, opened 2023-01-23
Make python example multiple lines so it's not massive :P

gewenyu99:
Uses the append pipe `>>` to break up examples.


gewenyu99:
https://user-images.githubusercontent.com/29069505/214127329-e034658e-0468-4735-84ed-6be370c69da2.mov

PR #135, opened 2023-01-24
Docs: Usage script readability

Meldiron:
This PR makes usage example scripts:

- More readable for humans
- Copyable from readme
- Copyable from terminal

### DIFF
--- a/runtimes/dotnet-3.1/README.md
+++ b/runtimes/dotnet-3.1/README.md
@@ -12,7 +12,17 @@ To learn more about runtimes, visit [Structure](https://github.com/open-runtimes
 
 ```bash
 mkdir dotnet-or && cd dotnet-or
-printf "using System.Threading.Tasks;\nusing System.Collections.Generic;\n\npublic async Task<RuntimeResponse> Main(RuntimeRequest req, RuntimeResponse res) => res.Json(new Dictionary<string, object>() {{ \"n\", new System.Random().NextDouble() }} );" > Index.cs
+tee -a Index.cs << END
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+public async Task<RuntimeResponse> Main(RuntimeRequest req, RuntimeResponse res)
+{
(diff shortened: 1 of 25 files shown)


Meldiron:
Tested all scripts manually

runtimes/dotnet-3.1/README.md
```diff
@@ -12,7 +12,17 @@ To learn more about runtimes, visit [Structure](https://github.com/open-runtimes
...
 ```bash
 mkdir dotnet-or && cd dotnet-or
-printf "using System.Threading.Tasks;\nusing System.Collections.Generic;\n\npublic async Task<RuntimeResponse> Main(RuntimeRequest req, RuntimeResponse res) => res.Json(new Dictionary<string, object>() {{ \"n\", new System.Random().NextDouble() }} );" > Index.cs
+tee -a Index.cs << END
+using System.Threading.Tasks;
+using System.Collections.Generic;
+
+public async Task<RuntimeResponse> Main(RuntimeRequest req, RuntimeResponse res)
+{
+    return res.Json(new Dictionary<string, object>() {{ "n", new System.Random().NextDouble() }} );
```
gewenyu99, line 21: We can probably format this better, now that we use `tee`

gewenyu99 (changes requested):
I'm good with this, 1 comment. Only concern is that you clearly explain what `tee` does in the instructions preceding this code block.

Meldiron:
@gewenyu99 above script we say `Add code into "Index.cs" file`. Do you think that is good enough or we should be more specific?

gewenyu99:
> @gewenyu99 above script we say `Add code into "Index.cs" file`. Do you think that is good enough or we should be more specific?

Good enough for now :)

gewenyu99 (approved):
LGTM

Meldiron:
This will be broken by V3. New branch should be started when v3 is merged.

Meldiron:
Closing in favour of: https://github.com/open-runtimes/open-runtimes/pull/154

PR #136, opened 2023-01-24
Open Runtimes V3

Meldiron:
# Open Runtimes V3

## Description

- Ensure brand. Slug should be `open-runtimes`, verbose should be `Open Runtimes`, and there is no shorthand (no `opr`). `openruntimes` should not be used.
- `req` and `res` params removed, `context` introduced
- `context.req` contains all information about http request
- `context.res` lets you configure all information about http request
- `res.json()` is now `context.res.json(). same for `res.send()` -> `context.res.send()`. New ones introduced, `.file()`, `.empty()` and `.redirect()`
- Multiple methods on response introduced such as `context.res.setHeader()`, `context.res.setBody()`, `context.res.setCode()`, as well as some helpers such as `context.res.setContentType()` or `context.res.setCookies()`
- `context.log()` and `context.error()` lets you reliably log. Native support (like`console.log` in Node) was removed 
- `req.variables` removed, system variables should be used instead (in Node, `process.env.MY_VAR`)
- `req.payload` renamed to `context.req.body`. Body is automatically parsed now to JSON if there is JSON content-type header. In future we plan to support more types such as XML, URL-encoded or Form Data bodies.
- Function now **has to** return. Instead of `res.json()`, now we do `return context.res.json()`
- Missing response now responds with 204 instead of causing timeout error
- Auth header renamed from `x-internal-challenge` to `x-open-runtimes-secret`
- Logs race condition issues are now solved
- **MUCH** better tests coverage

## Development

Runtimes:
- [x] Node
- [ ] (add others ...)

Tests coverage:

- [x] Logging to logs
- [x] Logging to error
- [x] Rrequest headers
- [x] Request headers MUST be ommiting `x-openruntimes-` ones
- [x] Request URL
- [x] Request method
- [x] Request body empty
- [x] Request body plaintext
- [x] Request body JSON (automatic parsing)
- [x] Internal challenge 500
- [x] Exception thrown
- [x] Double response
- [x] No response
- [x] Empty response
- [x] File response
- [x] Redirect response
- [x] JSON response
- [x] Text response
- [x] Emoji response
- [x] Headers response
- [x] Build-approach response
- [x] env variables
- [x] Old tests, if relevant

Docs:
- [ ] Ensure OPR is no longer used
- [ ] Update all examples and READMEs
- [ ] Update main readme
- [ ] Update opr/examples
- [ ] Update appwrite/function-starters


runtimes/cpp-17/.env
```diff
@@ -1,2 +1,4 @@
-INTERNAL_RUNTIME_KEY=secret-key
-INTERNAL_RUNTIME_ENTRYPOINT=index.cc
\ No newline at end of file
+OPEN_RUNTIMES_SECRET=secret-key
+OPEN_RUNTIMES_ENTRYPOINT=index.cc
+BUILD_COMMAND=""
```
eldadfux, line 3: Do we want an `OPEN_RUNTIMES` prefix for both `BUILD_COMMAND` and `START_COMMAND`?

runtimes/cpp-17/README.md
```diff
@@ -53,41 +74,42 @@ cd open-runtimes/runtimes/cpp-17
...
 }
 ```
 
 - To handle dependencies, you need to include a `CMakeLists.txt` file. Dependencies will be automatically cached and installed, so you don't need to include the `build` folder in your function.
 
-- The default entrypoint is `index.cc`. If your entrypoint differs, make sure to configure it using `INTERNAL_RUNTIME_ENTRYPOINT` environment variable, for instance, `INTERNAL_RUNTIME_ENTRYPOINT=src/app.cc`.
+- The default entrypoint is `index.cc`. If your entrypoint differs, make sure to configure it using `OPEN_RUNTIMES_ENTRYPOINT` environment variable during build, for instance, `OPEN_RUNTIMES_ENTRYPOINT=src/app.cc`.
 
 
 ## Authors
```
eldadfux, line 115: We should remove all authors.

eldadfux (commented):
@Meldiron great stuff! This is a game changing PR!

runtimes/cpp-17/README.md
Meldiron: Removed all authors

runtimes/cpp-17/.env
Meldiron: Good catch, it started as just test variable 😅 Updated.

eldadfux (approved):

PR #137, opened 2023-01-26
Feat: v3 python

Meldiron:
Python implementation of V3. Parent PR: https://github.com/open-runtimes/open-runtimes/pull/136

This PR also updates Node runtime to work with all headers lowercase


tests/resources/functions/python-3.10/tests.py
```diff
@@ -1,39 +1,68 @@
...
+    elif action == 'jsonResponse':
+        return context.res.json({ 'json': True, 'message': 'Developers are awesome.' })
+    elif action == 'redirectResponse':
+        return context.res.redirect('https://github.com/')
+    elif action == 'emptyResponse':
+        return context.res.empty()
+    elif action == 'noResponse':
+        context.res.send('This should be ignored, as it is not returned.')
+    elif action == 'doubleResponse':
+        context.res.send('This should be ignored.')
```
christyjacob4, line 18: Shouldn't this throw an error ( i.e. return a 500 response )

tests/resources/functions/node-18.0/tests.js
```diff
@@ -47,7 +54,7 @@ module.exports = async (context) => {
             context.log(4.2);
             context.log(true);
 
-            return context.res.send();
+            return context.res.send('');
```
christyjacob4, line 57: What is the behaviour of `res.send()` and `res.send('')`?

runtimes/python-3.10/server.py
```diff
@@ -1,98 +1,151 @@
...
+            'body': self._body,
+            'statusCode': self._statusCode,
+            'headers': self._headers,
+        }
 
-    def send(self, text, status = 200):
-        self._status = status
-        self._response = text
+    def json(self, obj, statusCode = 200, headers = {}):
+        headers['Content-Type'] = 'application/json'
```
christyjacob4, line 39: lower case headers ?

christyjacob4 (approved):
Looks good 👍

tests/resources/functions/node-18.0/tests.js
Meldiron: res.send should not be allowed and is not recommended. first argument is required.
res.send('') sends empty body, 200 status code, no headers.

I didnt notice this problem because JS allows anything. Python already complained that first argument is required

tests/resources/functions/python-3.10/tests.py
Meldiron: First one does not return, so its ignored. THat is the rest.

Double return is not allowed in genreal. language wont let you

runtimes/python-3.10/server.py
Meldiron: Good catch, updated.

Meldiron: `Location` fixed too

christyjacob4: Awesome 👍

tests/resources/functions/node-18.0/tests.js
christyjacob4: okay got it 👍

tests/resources/functions/python-3.10/tests.py
christyjacob4: okay

PR #140, opened 2023-01-28
Feat: v3 Ruby

Meldiron:
v3 implementation for Ruby. Parent PR:  https://github.com/open-runtimes/open-runtimes/pull/136


runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
 
 USER_CODE_PATH = '/usr/code-start';
 
-before do
-  content_type :json
-end
+class Response
```
abnegate, line 6: We should prefix the class names with `Runtime` for consistency across runtimes

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
...
+  def send(body, statusCode = 200, headers = {})
```
abnegate, line 7: Variables should be `snake_case` in Ruby, applied to whole PR

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
...
-    load(USER_CODE_PATH + '/' + ENV['INTERNAL_RUNTIME_ENTRYPOINT'])
-  rescue Exception => e
-    p e
-    status 500
-    return {stderr: e.backtrace.join("\n") }.to_json
+  request.env.each do |header, value|
+    if(header.start_with?('HTTP_'))
+      header = header[5..-1]
+      header = header.gsub("_", "-")
+      header = header.downcase
```
abnegate, line 152: You could chain these methods:

```suggestion
      header = header[5..-1]
        .gsub("_", "-")
        .downcase
```

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
...
 
-  def variables
-    @variables
+  attr_accessor :req
+  attr_accessor :res
+  attr_accessor :logs
+  attr_accessor :errors
+
+  def log(message)
+    if(message.kind_of?(Array) || message.kind_of?(Hash))
```
abnegate, line 72: It would be more Ruby-esque to omit the braces for `if` statements:

```suggestion
    if message.kind_of?(Array) || message.kind_of?(Hash)
```

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
...
+    end
   end
 
-  unless defined?(main = ())
-    status 500
-    return { stderr: 'File does not specify a main() function.' }.to_json
+  contentType = request.env['CONTENT_TYPE']
+  if(contentType == nil)
+    contentType = 'text/plain'
   end
```
abnegate, line 163: You could use a trailing if here:

```suggestion
content_type = 'text/plain' if content_type.nil?
```

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,271 @@
...
+  end
+
+  response.status = output['statusCode']
+  response.body = output['body']
+end
+
+get '*' do
+  handle(request, response)
+  return response
+end
```
abnegate, line 243: If you return the response from handle, then you can omit the separate return for each route:

```suggestion
  response.body = output['body']
  response
end

get '*' do
  return handle(request, response)
end
```

runtimes/ruby-3.1/server.rb
Meldiron: Updated.

runtimes/ruby-3.1/server.rb
Meldiron: Updated all

runtimes/ruby-3.1/server.rb
Meldiron: Fixed

runtimes/ruby-3.1/server.rb
Meldiron: Updated everywhere

runtimes/ruby-3.1/server.rb
Meldiron: Appleid here and in other places as well.

runtimes/ruby-3.1/server.rb
Meldiron: Applied

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,255 @@
...
+    end
+
     system_out = $stdout
     system_err = $stderr
-    user_out = StringIO.new
-    user_err = StringIO.new
-    $stdout = user_out
-    $stderr = user_err
+    customstd = StringIO.new
+    $stdout = customstd
```
christyjacob4, line 185: why the `$` ?

tests/resources/functions/ruby-3.1/tests.rb
```diff
@@ -1,36 +1,87 @@
...
+            'key2': key2,
+            'raw': context.req.body_string
+        })
+    when 'envVars'
+        return context.res.json({
+            'var': ENV['CUSTOM_ENV_VAR'] || nil,
+            'emptyVar': ENV['NOT_DEFINED_VAR'] || nil
+        })
+    when 'logs'
+        print('Native log')
```
christyjacob4, line 69: what happens if we use `puts` ? We should have a test case for that too

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,255 @@
...
+      @errors.push(message.to_s)
+    end
   end
 end
 
-class RuntimeResponse
-  def send(message)
-    @response = message
+def handle(request, response)
+  secret = request.env['HTTP_X_OPEN_RUNTIMES_SECRET'] || ''
```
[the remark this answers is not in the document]
Meldiron: From Sinatra, our HTTP server. https://stackoverflow.com/questions/25660827/accessing-headers-from-sinatra

runtimes/ruby-3.1/example/index.rb
```diff
@@ -1,25 +1,13 @@
...
-    payload = JSON.parse(req.payload)
+def main(context)
+    payload = context.req.body
 
     todo = JSON.parse(HTTParty.get('https://jsonplaceholder.typicode.com/todos/' + (payload['id'] || '1')).body)
 
-    return res.json({
-        :message => 'Hello Open Runtimes 👋',
-        :todo => todo
+    return context.res.json({
```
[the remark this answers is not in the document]
Meldiron: Yes. Example here: https://github.com/open-runtimes/open-runtimes/blob/7f2193ba528b3e41de35e940c2c27d10d8348a84/tests/resources/functions/ruby-3.1/tests.rb#L65

tests/resources/functions/ruby-3.1/tests.rb
Meldiron: I tried it manually and it works fine.
Do we really need test for it? In JS we would need like 5 more tests then, for all log types.

It would make tests inconsistent across languages and would need runtime-specific tests. Which I would strongly want to avoid.

abnegate: `puts` is more idiomatic Ruby, if anything I would test for that instead of `print`.

I would check a dump of the entire log too instead of only checking if it contains a string, I suspect you'll have extra newlines when using `puts` as it adds a newline, and print does not

Meldiron: Updates test to use `puts`

runtimes/ruby-3.1/server.rb
```diff
@@ -3,117 +3,255 @@
...
+  output['headers'] = {} if output['headers'].nil?
+
+  output['headers'].each do |header, value|
+    if !header.downcase.start_with?('x-open-runtimes-')
+      response.headers[header.downcase] = value
+    end
+  end
+
+  if customstd.string != nil && !customstd.string.empty?
+    context.log('Unsupported log noticed. Use context.log() or context.error() for logging.')
```
PineappleIOnic, line 215: Is "noticed" the best wording here? Wouldn't "detected" work better?

tests/resources/functions/ruby-3.1/tests.rb
```diff
@@ -1,36 +1,87 @@
...
+        context.res.send('This should be ignored, as it is not returned.')
+        return nil
+    when 'doubleResponse'
+        context.res.send('This should be ignored.')
+        return context.res.send('This should be returned.')
+    when 'headersResponse'
+        return context.res.send('OK', 200, {
+            'first-header': 'first-value',
+            'second-header': context.req.headers['x-open-runtimes-custom-in-header'] || 'missing',
+            'x-open-runtimes-custom-out-header': 'third-value'
```
PineappleIOnic, line 26: What's the difference between first-header, second-header and x-open-runtime-custom-out-header? That third one is quite different from the others

tests/resources/functions/ruby-3.1/tests.rb
```diff
@@ -1,36 +1,87 @@
...
+
+        context.log({ 'objectKey': 'objectValue' })
+        context.log([ 'arrayValue' ])
+
+        return context.res.send('')
+    when 'library'
+        todo = JSON.parse(HTTParty.get('https://jsonplaceholder.typicode.com/todos/' + context.req.body_string).body)
+        return context.res.json({ 'todo': todo })
+    else
+        raise 'Unkonwn action'
```
PineappleIOnic, line 85:
```suggestion
        raise 'Unknown action'
```

PineappleIOnic (changes requested):

tests/resources/functions/ruby-3.1/tests.rb
Meldiron: First is simply checking if response headers works.
Second is checking if we can read request headers.
Third is checking if reserved keyword (x-open-runtimes) headers are missing from response - intended.

runtimes/ruby-3.1/server.rb
Meldiron: Using detected instead.

tests/resources/functions/ruby-3.1/tests.rb
Meldiron: Updated.

tests/resources/functions/ruby-3.1/tests.rb
PineappleIOnic: Alright, gotcha

PineappleIOnic (approved):

PR #141, opened 2023-01-29
Feat: v3 Dart

Meldiron:
Dart implementation of V3. Parent PR: https://github.com/open-runtimes/open-runtimes/pull/136


runtimes/dart-2.18/function_types.dart
```diff
@@ -1,33 +1,77 @@
...
+  dynamic body;
+  Map<String, dynamic> headers;
+  String method;
+  String url;
+  String path;
+  int port;
+  String scheme;
+  String host;
+  String queryString;
+  Map<String, String> query;
```
lohanidamodar, line 14: Why don't we use `URI` instead of this? `URI` is not supported in any other language?

runtimes/dart-2.18/server.dart
```diff
@@ -10,56 +10,121 @@ import 'function_types.dart';
...
+          output = await user_code.handle(context);
         },
-        (e, stackTrace) => print('$e $stackTrace'),
         zoneSpecification: ZoneSpecification(
           print: (Zone self, ZoneDelegate parent, Zone zone, String line) {
-            userLogs.add(line);
+            customstd += line;
           },
         ),
       );
```
lohanidamodar, line 96: if we remove support for `console.log` in node, here as well we can remove support for `print` overwriting, so we can remove everything inside try and simplify it
```suggestion
          output = await user_code.handle(context);
```

lohanidamodar (approved):
Looks good, left few queries

runtimes/dart-2.18/server.dart
Meldiron: We are keeping console.log logic.

When you use natice way to log, we capture it. Instead of logging the message, we write a log saying "Please use context.log instead".

This way devs who use native log for first time (natural behaviour) will see exactly what they need to do to log properly.

runtimes/dart-2.18/function_types.dart
Meldiron: We decided to not use URI component and stay consistent across all runtimes.

runtimes/dart-2.18/server.dart
```diff
@@ -10,56 +10,121 @@ import 'function_types.dart';
...
+      if(!bodyString.isEmpty) {
+        body = jsonDecode(bodyString);
+      } else {
+        body = {};
+      }
+    }
+
+    String hostHeader = req.headers['host'] ?? '';
+
+    String path = '/' + req.url.path;
```
stnguyen90, line 40: I think we should be using `req.requestedUri` instead. debugger shows the expected values in `req.requestedUri`:

image

runtimes/dart-2.18/server.dart
```diff
@@ -10,56 +10,121 @@ import 'function_types.dart';
...
+    for(final param in queryString.split("&")) {
+      final parts = param.split("=");
+
+      final key = parts[0];
+      final value = parts.sublist(1).join('=');
+
+      if(key != null && !key.isEmpty) {
+        query[key] = value;
+      }
+    }
```
stnguyen90, line 65: I don't think we should manually parse the query string like this because we might not catch some edge cases with & and =. Also, this doesn't quite handle arrays well.

runtimes/dart-2.18/function_types.dart
```diff
@@ -1,33 +1,77 @@
...
+    }
   }
 
-  Response json(Map<String, dynamic> json, {int status = 200}) {
-    _text = json;
-    _status = status;
-    return this;
+  void error(dynamic message) {
+    if (message is List || message is Map) { 
+      this.errors.add(jsonEncode(message));
```
stnguyen90, line 72: This could throw an exception if it's not possible to jsonEncode message.

stnguyen90 (commented):
Here are my preliminary comments. I'd really like to test this a bit more, though.

runtimes/dart-2.18/server.dart
```diff
@@ -10,56 +10,121 @@ import 'function_types.dart';
...
+
+    Request contextReq = new Request(bodyString: bodyString, body: body, headers: headers, method: method, url: url, path: path, port: port, scheme: scheme, host: host, queryString: queryString, query: query);
+    Response contextRes = new Response();
+    Context context = new Context(contextReq, contextRes);
+
+    String customstd = "";
+
+    dynamic output = null;
+    try {
+      await runZoned(
```
stnguyen90, line 87: Another downside about using this `runZoned()` is the stacktrace ends up including the zone:

```
Exception: My custom error
#0      main (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/example/lib/main.dart:6:3)
#1      main.<anonymous closure>.<anonymous closure> (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/server.dart:107:36)
#2      _rootRun (dart:async/zone.dart:1391:13)
#3      _CustomZone.run (dart:async/zone.dart:1293:19)
#4      _runZoned (dart:async/zone.dart:1829:10)
#5      runZoned (dart:async/zone.dart:1751:10)
#6      main.<anonymous closure> (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/server.dart:105:13)
<asynchronous suspension>
#7      handleRequest (package:shelf/shelf_io.dart:137:16)
<asynchronous suspension>
```

runtimes/dart-2.18/function_types.dart
stnguyen90: For example, if my function was this:

```dart
import 'dart:async';

class MyObject {
  String name;
  int age;
  MyObject({required this.name, required this.age});
}

Future<dynamic> main(final context) async {
  context.log([MyObject(name: 'name', age: 12)]);
  return context.res.json({
    'message': "Hello Open Runtimes 👋",
  });
}
```

An exception would be thrown and the `x-open-runtimes-errors` header would have:

```
Converting object to an encodable object failed: Instance of 'MyObject'
#0      _JsonStringifier.writeObject (dart:convert/json.dart:794:7)
#1      _JsonStringifier.writeList (dart:convert/json.dart:842:7)
#2      _JsonStringifier.writeJsonValue (dart:convert/json.dart:824:7)
#3      _JsonStringifier.writeObject (dart:convert/json.dart:785:9)
#4      _JsonStringStringifier.printOn (dart:convert/json.dart:983:17)
#5      _JsonStringStringifier.stringify (dart:convert/json.dart:968:5)
#6      JsonEncoder.convert (dart:convert/json.dart:345:30)
#7      JsonCodec.encode (dart:convert/json.dart:231:45)
#8      jsonEncode (dart:convert/json.dart:114:10)
#9      Context.log (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/function_types.dart:64:21)
#10     main (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/example/lib/main.dart:10:11)
#11     main.<anonymous closure>.<anonymous closure> (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/server.dart:107:36)
#12     _rootRun (dart:async/zone.dart:1391:13)
#13     _CustomZone.run (dart:async/zone.dart:1293:19)
#14     _runZoned (dart:async/zone.dart:1829:10)
#15     runZoned (dart:async/zone.dart:1751:10)
#16     main.<anonymous closure> (file:///Users/steven/Projects/open-runtimes-open-runtimes/runtimes/dart-2.18/server.dart:105:13)
<asynchronous suspension>
#17     handleRequest (package:shelf/shelf_io.dart:137:16)
<asynchronous suspension>
```

runtimes/dart-2.18/server.dart
stnguyen90: > This way devs who use native log for first time (natural behaviour) will see exactly what they need to do to log properly.

I don't quite agree with this approach. rewriting standard functionality has caused many problems in our current version of Appwrite. Sometimes, people do want to see the output in the docker logs.

runtimes/dart-2.18/server.dart
```diff
@@ -10,56 +10,121 @@ import 'function_types.dart';
...
+          output = await user_code.main(context);
         },
-        (e, stackTrace) => print('$e $stackTrace'),
         zoneSpecification: ZoneSpecification(
           print: (Zone self, ZoneDelegate parent, Zone zone, String line) {
-            userLogs.add(line);
+            customstd += line;
           },
         ),
       );
```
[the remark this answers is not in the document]
Meldiron: We are, and if we spot that you used it, we will ignore what you printed, and print "Use context.log()" instead. It helps devs quickly figure out why native way does not work.

runtimes/dart-2.18/server.dart
Meldiron: Is that a problem? We use runZoned to override print method.

runtimes/dart-2.18/function_types.dart
Meldiron: Hmm what do you suggest? Wrap it into try/catch, and do toString as fallback?

runtimes/dart-2.18/server.dart
Meldiron: Hmm we currently dont support arrays (`?key[]=a&key[]=b`), as well as json-like structure (`?key[name]=a&key[role]=b`).
Tho this could be problematic in some strictily-typed languages, as value value be `string`, `string[]`, or `{ string: string }`

Meldiron: Also not sure if it's possible to entirely avoid this in all runtimes. I would imagine that C++ doesnt have a native way to do this, and would require a library, or a solution just this one.

runtimes/dart-2.18/server.dart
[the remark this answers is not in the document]
stnguyen90: I really don't think this is a good idea. This will make it impossible for people to use use the standard logs if they want to.

runtimes/dart-2.18/function_types.dart
stnguyen90: That sounds reasonable

runtimes/dart-2.18/server.dart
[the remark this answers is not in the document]
Meldiron: This might be topic where someone will always complain, no matter which way we go.
I would prefer goign with this solution as it forces developer to start using `context.log`, and provide us any feedback if this method miss-behaves.

If we get too many complains, we can remove this logic in future.

stnguyen90 (approved):

runtimes/dart-2.18/server.dart
obiwanzenobi: https://github.com/open-runtimes/open-runtimes/issues/174
probably this is the problem you mentioned - and indeed runZoned with try catch above should solve this bug

PR #144, opened 2023-01-30
Feat: v3 .NET

Meldiron:
Implementation of V3 in dotnet. Parent PR: https://github.com/open-runtimes/open-runtimes/pull/136


runtimes/dotnet-6.0/Example/Index.cs
```diff
@@ -1,24 +0,0 @@
-using Newtonsoft.Json;
```
abnegate, line 1: This should be updated instead of removed

runtimes/dotnet-6.0/build.sh
```diff
@@ -28,16 +28,6 @@ build() {
...
-
-    case ${INTERNAL_RUNTIME_ENTRYPOINT##*.} in 
-        cs) write_cs_wrapper ;;
-        fs) write_fs_wrapper ;;
-        vb) write_vb_wrapper ;;
-    esac
-
-    # Remove the user code file (copy)
-    rm "${INTERNAL_RUNTIME_ENTRYPOINT}"
-
```
abnegate, line 40: How is this working with all of this removed 🤔

runtimes/dotnet-6.0/src/Program.cs
```diff
@@ -3,69 +3,163 @@
...
+    string Secret = Request.Headers.TryGetValue("x-open-runtimes-secret", out var SecretValue) ? (string) SecretValue : "";
+    if(Secret == "" || Secret != (Environment.GetEnvironmentVariable("OPEN_RUNTIMES_SECRET") ?? ""))
     {
-        return Results.Problem(
-            detail: "Unauthorized",
-            statusCode: 500);
+        return new CustomResponse("Unauthorized. Provide correct \"x-open-runtimes-secret\" header.", 500);
     }
 
+    StreamReader Reader = new StreamReader(Request.Body);
```
abnegate, line 22: Let's use `var` instead of specific types for consistency

runtimes/dotnet-6.0/src/Program.cs
```diff
@@ -3,69 +3,163 @@
...
 {
-    if (challengeId is not null 
-        && challengeId != Environment.GetEnvironmentVariable("INTERNAL_RUNTIME_KEY"))
+    string Secret = Request.Headers.TryGetValue("x-open-runtimes-secret", out var SecretValue) ? (string) SecretValue : "";
+    if(Secret == "" || Secret != (Environment.GetEnvironmentVariable("OPEN_RUNTIMES_SECRET") ?? ""))
     {
-        return Results.Problem(
-            detail: "Unauthorized",
-            statusCode: 500);
+        return new CustomResponse("Unauthorized. Provide correct \"x-open-runtimes-secret\" header.", 500);
```
abnegate, line 19: Can we use the built-in `Results` instead of a `CustomResponse` class?

runtimes/dotnet-6.0/src/Program.cs
```diff
@@ -3,69 +3,163 @@
...
+            Body = JsonConvert.DeserializeObject<Dictionary<string, object>>(BodyString);
+        }
+    }
+
+    if(Body == null)
+    {
+        Body = new Dictionary<string, object>();
+    }
+
+    String HostHeader = Request.Headers.TryGetValue("host", out var HostHeaderValue) ? (string) HostHeaderValue : "";
```
abnegate, line 57: You can use `out string` to avoid the cast

runtimes/dotnet-6.0/src/Program.cs
```diff
@@ -3,69 +3,163 @@
...
+
+    string ContentType = Request.Headers.TryGetValue("content-type", out var ContentTypeValue) ? (string) ContentTypeValue : "";
+    if(ContentType.Contains("application/json"))
+    {
+        if(String.IsNullOrEmpty(BodyString))
+        {
+            Body = new Dictionary<string, object>();
+        } else
+        {
+            Body = JsonConvert.DeserializeObject<Dictionary<string, object>>(BodyString);
```
abnegate, line 48: Let's use `JsonSerializer` instead of `JsonConvert` to remove the dependency on `Newtonsoft.Json`

runtimes/dotnet-6.0/src/RuntimeContext.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Text.Json;
+
+namespace DotNetRuntime
+{
+	public class RuntimeContext
+	{
+		public RuntimeRequest Req { get; set; }
+		public RuntimeResponse Res { get; set; }
+
+		public ArrayList Logs = new ArrayList();
```
abnegate, line 11: Let's use `List<string>` instead of `ArrayList`

runtimes/dotnet-6.0/src/RuntimeRequest.cs
```diff
@@ -2,13 +2,32 @@ namespace DotNetRuntime
...
+		public string QueryString { get; set; }
+		public Dictionary<string, string> Query { get; set; }
+		public int Port { get; set; }
 
-		public Dictionary<string, string> Variables { get; set; } =
-			new Dictionary<string, string>();
-
-		public Dictionary<string, string> Headers { get; set; } =
-			new Dictionary<string, string>();
+		public RuntimeRequest(string BodyString, object Body, Dictionary<string, string> Headers, string Method, string Url, string Host, string Scheme, string Path, string QueryString, Dictionary<string, string> Query, int Port)
```
abnegate, line 17: Let's be consistent across all runtimes with param order, following how they appear in an actual request: method, scheme, host, port, path, query, headers, body

tests/resources/functions/dotnet-6.0/Tests.cs
```diff
@@ -1,46 +1,128 @@
...
+                Context.Log(Obj);
+
+                ArrayList Arr = new ArrayList();
+                Arr.Add("arrayValue");
+
+                Context.Log(Arr);
+
+                return Context.Res.Send("");
+            case "library":
+                var response = await http.GetStringAsync($"https://jsonplaceholder.typicode.com/todos/" + Context.Req.BodyString);
```
abnegate, line 117: You can use interpolation with `$` strings instead of using + like `$"https...{context.Req.BodyString}"`

runtimes/dotnet-6.0/build.sh
Meldiron: Developer must do this in their code now. No more magic. All documented in examples and notes in README.md

runtimes/dotnet-6.0/src/Program.cs
Meldiron: I wasnt able to find a way to use Results with custom body, status code and headers. That's why Implemented this one.

runtimes/dotnet-6.0/Example/Index.cs
Meldiron: It was moved to `example` folder (lower-case) for consistency with other languages.

runtimes/dotnet-6.0/src/Program.cs
Meldiron: Updated all.

PR #148, opened 2023-02-03
Fix: Node & Python

Meldiron:
Fix node and python v3 to support new tests

### DIFF
--- a/runtimes/node-18.0/server.js
+++ b/runtimes/node-18.0/server.js
@@ -34,13 +34,35 @@ const server = micro(async (req, res) => {
         headers[header.toLowerCase()] = req.headers[header];
     });
 
+    const scheme = (req.headers['x-forwarded-proto'] ?? '') === 'https' ? 'https' : 'http';
+    const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
+    const path = req.url.includes('?') ? req.url.split('?')[0] : req.url;
+    const queryString = req.url.includes('?') ? req.url.split('?')[1] : '';
+    const query = {};
+    for(const param of queryString.split('&')) {
+        const [ key, value ] = param.split('=');
+
+        if(key) {
+            query[key] = value;
+        }
+    }
+
+    const url = `${scheme}://${host}${port === 80 ? '' : `:${port}`}${path}${queryString === '' ? '' : `?${queryString}`}`;
+
     const context = {
         req: {
             bodyString,
             body,
             headers,
             method: req.method,
-            url: req.url
+            host,
+            scheme,
+            query,
+            queryString,
+            port,
+            url,
+            path
         },
         res: {
             send: function (body, statusCode = 200, headers = {}) {
@@ -70,14 +92,14 @@ const server = micro(async (req, res) => {
             if (message instanceof Object || Array.isArray(message)) {
                 logs.push(JSON.stringify(message));
             } else {
-                logs.push(message.toString());
+                logs.push(message + "");
             }
         },
         error: function (message) {
             if (message instanceof Object || Array.isArray(message)) {
                 errors.push(JSON.stringify(message));
             } else {
-                errors.push(message.toString());
+                errors.push(message + "");
             }
         },
     };
--- a/runtimes/node-18.0/server.js
+++ b/runtimes/node-18.0/server.js
@@ -31,8 +31,9 @@ const server = micro(async (req, res) => {
     });
 
     const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
+    const defaultPort = scheme === 'https' ? '443' : '80';
     const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
-    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : defaultPort);
     const path = req.url.includes('?') ? req.url.split('?')[0] : req.url;
     const queryString = req.url.includes('?') ? req.url.split('?')[1] : '';
     const query = {};
@@ -45,7 +46,7 @@ const server = micro(async (req, res) => {
         }
     }
 
-    const url = `${scheme}://${host}${port === 80 ? '' : `:${port}`}${path}${queryString === '' ? '' : `?${queryString}`}`;
+    const url = `${scheme}://${host}${port.toString() === defaultPort ? '' : `:${port}`}${path}${queryString === '' ? '' : `?${queryString}`}`;
 
     const context = {
         req: {
--- a/runtimes/node-18.0/server.js
+++ b/runtimes/node-18.0/server.js
@@ -22,11 +22,7 @@ const server = micro(async (req, res) => {
     let body = bodyString;
 
     if (contentType.includes('application/json')) {
-        if(bodyString) {
-            body = await parseJson(req);
-        } else {
-            body = {};
-        }
+        body = await parseJson(req);
     }
 
     const headers = {};
--- a/runtimes/node-18.0/server.js
+++ b/runtimes/node-18.0/server.js
@@ -5,7 +5,7 @@ const { text: parseText, json: parseJson, buffer: parseBuffer, send } = require(
 const USER_CODE_PATH = '/usr/code-start';
 
 const server = micro(async (req, res) => {
-    if ((req.headers[`x-open-runtimes-secret`] ?? '') !== process.env['OPEN_RUNTIMES_SECRET']) {
+    if ((req.headers[`x-open-runtimes-secret`] ?? '') !== (process.env['OPEN_RUNTIMES_SECRET'] ?? '')) {
         return send(res, 500, 'Unauthorized. Provide correct "x-open-runtimes-secret" header.');
     }
 
--- a/runtimes/node-18.0/server.js
+++ b/runtimes/node-18.0/server.js
@@ -34,33 +34,9 @@ const server = micro(async (req, res) => {
         headers[header.toLowerCase()] = req.headers[header];
     });
 
-    let host = "";
-    let port = 80;
-    let user = "";
-    let password = "";
-
-    const hostHeader = req.headers['host'] ?? '';
-
-    const authHeader = hostHeader.includes('@') ? hostHeader.split('@')[0] : '';
-    const hostnameHeader = hostHeader.includes('@') ? hostHeader.split('@')[1] : hostHeader;
-
-    if(authHeader.includes(':')) {
-        user = authHeader.split(':')[0];
-        password = authHeader.split(':')[1];
-    } else {
-        user = authHeader;
-        password = "";
-    }
-
-    if(hostnameHeader.includes(':')) {
-        host = hostnameHeader.split(':')[0];
-        port = +(hostnameHeader.split(':')[1]);
-    } else {
-        host = hostnameHeader;
-        port = 80;
-    }
-
     const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
+    const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
     const path = req.url.includes('?') ? req.url.split('?')[0] : req.url;
     const queryString = req.url.includes('?') ? req.url.split('?')[1] : '';
     const query = {};
@@ -86,9 +62,7 @@ const server = micro(async (req, res) => {
             queryString,
             port,
             url,
-            path,
-            user,
-            password
+            path
         },
         res: {
             send: function (body, statusCode = 200, headers = {}) {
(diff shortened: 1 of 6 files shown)


runtimes/node-18.0/server.js
```diff
@@ -5,7 +5,7 @@ const { text: parseText, json: parseJson, buffer: parseBuffer, send } = require(
 const USER_CODE_PATH = '/usr/code-start';
 
 const server = micro(async (req, res) => {
-    if (req.headers[`x-open-runtimes-secret`] !== process.env['OPEN_RUNTIMES_SECRET']) {
+    if ((req.headers[`x-open-runtimes-secret`] ?? '') !== process.env['OPEN_RUNTIMES_SECRET']) {
```
Meldiron, line 8:
```suggestion
    if ((req.headers[`x-open-runtimes-secret`] ?? '') !== (process.env['OPEN_RUNTIMES_SECRET'] ?? '')) {
```

runtimes/python-3.10/server.py
```diff
@@ -66,39 +72,70 @@ def __init__(self):
...
+        else:
+            self._errors.append(str(message))
 
 HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 
 @app.route('/', defaults={'u_path': ''}, methods = HTTP_METHODS)
 @app.route('/<path:u_path>', methods = HTTP_METHODS)
 def handler(u_path):
-    if (request.headers.get('x-open-runtimes-secret') != os.getenv('OPEN_RUNTIMES_SECRET')):
+    if (request.headers.get('x-open-runtimes-secret', '') != os.getenv('OPEN_RUNTIMES_SECRET')):
```
Meldiron, line 92:
```suggestion
    if (request.headers.get('x-open-runtimes-secret', '') != os.getenv('OPEN_RUNTIMES_SECRET', '')):
```

runtimes/node-18.0/server.js
```diff
@@ -18,25 +18,51 @@ const server = micro(async (req, res) => {
...
+            body = {};
+        }
     }
 
     const headers = {};
     Object.keys(req.headers).filter((header) => !header.toLowerCase().startsWith('x-open-runtimes-')).forEach((header) => {
         headers[header.toLowerCase()] = req.headers[header];
     });
 
+    const scheme = (req.headers['x-forwarded-proto'] ?? '') === 'https' ? 'https' : 'http';
```
Meldiron, line 37:
```suggestion
    const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
```

runtimes/node-18.0/server.js
```diff
@@ -5,7 +5,7 @@ const { text: parseText, json: parseJson, buffer: parseBuffer, send } = require(
 const USER_CODE_PATH = '/usr/code-start';
 
 const server = micro(async (req, res) => {
-    if (req.headers[`x-open-runtimes-secret`] !== process.env['OPEN_RUNTIMES_SECRET']) {
+    if ((req.headers[`x-open-runtimes-secret`] ?? '') !== (process.env['OPEN_RUNTIMES_SECRET'] ?? '')) {
```
christyjacob4, line 8: uff this is dangerous 👀 

If we accidentally forget to set the secret, anyone can bypass the this check by not passing the `x-open-runtimes-secret` header

Meldiron: We can set default value in Dockerfile. That way developer would need to purposely set it to empty string.
I find it valuable to give this option. Let's say you are running public static hosting. It gives value being able to just visit URL and see website, without having to set header (which browsers dont do)

runtimes/node-18.0/server.js
```diff
@@ -18,25 +18,51 @@ const server = micro(async (req, res) => {
...
     }
 
     const headers = {};
     Object.keys(req.headers).filter((header) => !header.toLowerCase().startsWith('x-open-runtimes-')).forEach((header) => {
         headers[header.toLowerCase()] = req.headers[header];
     });
 
+    const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
+    const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
```
christyjacob4, line 39: whats the + supposed to do ?

Meldiron: cast from string to integer.

runtimes/node-18.0/server.js
```diff
@@ -18,25 +18,51 @@ const server = micro(async (req, res) => {
...
+    const path = req.url.includes('?') ? req.url.split('?')[0] : req.url;
+    const queryString = req.url.includes('?') ? req.url.split('?')[1] : '';
```
christyjacob4, line 41: why are we doing all these steps manually? We can use the URL class of Node to do all this 

Screenshot 2023-02-07 at 3 42 55 PM

runtimes/node-18.0/server.js
```diff
@@ -62,30 +88,28 @@ const server = micro(async (req, res) => {
...
             }
-            errors.push(args.join(" "));
         },
     };
 
+    console.stdlog = console.log.bind(console);
+    console.stderror = console.error.bind(console);
+    console.stdinfo = console.info.bind(console);
+    console.stddebug = console.debug.bind(console);
+    console.stdwarn = console.warn.bind(console);
```
christyjacob4, line 111: why are we still binding this ? haven't we stopped using it ?

runtimes/python-3.10/server.py
```diff
@@ -47,58 +47,95 @@ def redirect(self, url, statusCode = 301, headers = {}):
...
+    for param in context.req.queryString.split('&'):
+        pair = param.split('=')
+
+        if pair[0]:
+            context.req.query[pair[0]] = pair[1]
+
+    host = request.headers.get('host', '')
+    if ':' in host:
+        context.req.host = host.split(':')[0]
+        context.req.port = int(host.split(':')[1])
```
christyjacob4, line 118: why do you need to use string splitting when you have the parsed URL already ?

runtimes/node-18.0/server.js
Meldiron: Developers will always use console.log. We still bind so we can see if they used console.log. If they did, we tell them (in log) that "This is not supported, please use context.log() instad".

runtimes/python-3.10/server.py
Meldiron: host header could be a domain like `myapp.com`, or it could be IP with port like `12.44.245.12:3000`. We split to separate port and host from this header.

runtimes/node-18.0/server.js
christyjacob4: If unit tests can cover all possible scenarios, i'm okay with it. Else its safest to go with a standard library when it exists. They're battle tested over the years

runtimes/node-18.0/server.js
christyjacob4: okay 👍

runtimes/python-3.10/server.py
christyjacob4: No I was not referring specifically to the host & port. I was referring to the manual parsing of strings as opposed to using the standard library

runtimes/node-18.0/server.js
christyjacob4: okay 👍 . This can be avoided if we use the std library

runtimes/node-18.0/server.js
christyjacob4:
```suggestion
     // This will check for the case when secret is undefined 
    if (!process.env['OPEN_RUNTIMES_SECRET']) {
        return send(res, 500, 'Unauthorized. Provide correct "x-open-runtimes-secret" header.');
    }
    // This will check all the other cases
    if (req.headers[`x-open-runtimes-secret`] !== process.env['OPEN_RUNTIMES_SECRET']) {
        return send(res, 500, 'Unauthorized. Provide correct "x-open-runtimes-secret" header.');
    }
```

Also are the 500 errors displayed in the function logs in the console ? If yes, we need to change the error message to not expose the `x-open-runtimes-secret` header. Its something internal to Appwrite and end-developers need not see this message in the UI. It should be logged to sentry though

runtimes/node-18.0/server.js
Meldiron: Ill use std library.

runtimes/node-18.0/server.js
Meldiron: Ill use std library.

runtimes/python-3.10/server.py
Meldiron: Ill use std library.

runtimes/node-18.0/server.js
Meldiron: I dont agree with not allowing empty secret. To compare, it's like not allowing guest to execute Appwrite function.
Anyway, Ill implement this error and add a test for it.

runtimes/node-18.0/server.js
```diff
@@ -30,13 +30,35 @@ const server = micro(async (req, res) => {
...
     });
 
+    const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
+    const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
+    const path = req.url.includes('?') ? req.url.split('?')[0] : req.url;
+    const queryString = req.url.includes('?') ? req.url.split('?')[1] : '';
+    const query = {};
+    for(const param of queryString.split('&')) {
+        const [ key, value ] = param.split('=');
```
abnegate, line 40: This will drop any part of a query param value containing an `=`, which is valid. E.g. `id=1234=abcd` will parse as `{ id: "1234"}`

runtimes/node-18.0/server.js
```diff
@@ -30,13 +30,35 @@ const server = micro(async (req, res) => {
         headers[header.toLowerCase()] = req.headers[header];
     });
 
+    const scheme = (req.headers['x-forwarded-proto'] ?? 'http');
+    const host = req.headers['host'].includes(':') ? req.headers['host'].split(':')[0] : req.headers['host'];
+    const port = +(req.headers['host'].includes(':') ? req.headers['host'].split(':')[1] : '80');
```
abnegate, line 35: Should we always assume `http` and port `80` if none are found?

Meldiron: There has to be a protocol, and there has to be a port. those sound as reasonable defaults.

> If the port is empty or not given, port 80 is assumed (HTTP RFC)

runtimes/node-18.0/server.js
Meldiron: Interesting.. Same for `&` I guess, I will investigate if those symbols are allowed and what is expected behaviour

Meldiron: Updated, added new test

runtimes/node-18.0/server.js
abnegate: I think we should use 443 as the default if the scheme is `https`

runtimes/node-18.0/server.js
```diff
@@ -30,13 +30,36 @@ const server = micro(async (req, res) => {
...
+    for(const param of queryString.split('&')) {
+        let [key, ...valueArr] = param.split('=');
+        const value = valueArr.join('=');
+
+        if(key) {
+            query[key] = value ?? '';
+        }
+    }
+
+    const url = `${scheme}://${host}${port === 80 ? '' : `:${port}`}${path}${queryString === '' ? '' : `?${queryString}`}`;
```
abnegate, line 48: If scheme is `https` and port `443` we should not include the port

Meldiron: Fixed, updated tests.

runtimes/node-18.0/server.js
Meldiron: Fixed, updated tests.

abnegate (commented):
Let's add a test for logging object instances like `new Class()` etc, will be important in other languages

abnegate (approved):

runtimes/python-3.10/server.py
```diff
@@ -47,58 +47,97 @@ def redirect(self, url, statusCode = 301, headers = {}):
...
+        else:
+            self.errors.append(str(message))
 
 HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
 
 @app.route('/', defaults={'u_path': ''}, methods = HTTP_METHODS)
 @app.route('/<path:u_path>', methods = HTTP_METHODS)
 def handler(u_path):
-    if (request.headers.get('x-open-runtimes-secret') != os.getenv('OPEN_RUNTIMES_SECRET')):
+    if (request.headers.get('x-open-runtimes-secret', '') != os.getenv('OPEN_RUNTIMES_SECRET', '')):
```
christyjacob4, line 92: This condition will be `false` when both the header and env var is empty.

In which case the security check is bypasses

christyjacob4 (changes requested):
Left one comment. Rest looks good 🆗

runtimes/python-3.10/server.py
Meldiron: Added check for empty/missing secret header to always return 500.

Issue #158, opened 2023-04-11
🚀 Feature: Adding Bun runtime


byawitz:
###  🔖 Feature description
Adding [Bun](https://bun.sh/) to open-runtimes.

 ###  🎤 Pitch
Bun is a **real fast** JavaScript runtime.

Bun advantages:
- Built-in [server](https://bun.sh/docs/api/http#start-a-server-bun-serve)
- Use typescript files - no need to transpile it happens on the fly.
- Faster loading time.
- Blazing fast request processing.
- Binary lock file.

I've created a few projects with Bun, and the process - building, testing, and producing - was fast and efficient. 

 ###  📝 Developemnt 
I can chip in and create the Bun open runtime.

👀 Have you spent some time checking if this issue has been raised before?
✅ I checked and didn't find a similar issue

🏢 Have you read the Code of Conduct?
✅ I have read the [Code of Conduct](https://github.com/open-runtimes/open-runtimes/blob/main/CODE_OF_CONDUCT.md)

joeyouss:
Thank you for this feature idea!

Looping in @Meldiron for feedback on this

byawitz:
I've just run a few benchmarks comparing the results of the same code in Bun vs. Node19 runtime.

The results are implacable; a table here represents the results of the K6 benchmark. _(thanks to @Meldiron for that one)_

| Paramter | Bun | Node  |
| --- | --- | --- | 
| Max VUS | 5,000 | 5,000 | 
| http_req_connecting | 1.47µs | 2.58µs | 
| http_req_duration | 60.07ms | 177.12ms | 
|  http_req_waiting | 60.32ms |  177.11ms | 
|  http_reqs | 2,499,396 | 857,188 | 
| Per second | 27,769 | 9,523 |

Both tests ran for 90 seconds using a Stress testing method.
```javascript
export const options = {
 stages: [
   {duration: '30s', target: 20},
   {duration: '10s', target: 5000},
   {duration: '20s', target: 10},
   {duration: '10s', target: 5000},
   {duration: '20s', target: 0},
 ],
};
```

I've uploaded a pull request of the Bun runtime in a ready-to-use state to be checked when ready.

Thanks

joeyouss:
Thank you for raising the PR. We will get it reviewed ASAP

PR #161, opened 2023-04-20
Feat: Bun runtime 0.5.9

byawitz:
Feature for #158


byawitz:
There is a small bug in the Teat typescript file. Which I will fix tomorrow.

PR #171, opened 2023-07-06
[Node] - Adapting servers to pass the SIGINT to the child server.

byawitz:
### start.sh
Inside each of the `start.sh` files, I've added a [trap](https://www.geeksforgeeks.org/shell-scripting-bash-trap-command/) to catch `SIGTERM` requests, and it's needed to kill the `node /usr/local/src/server.js` command (that being handled by **pm2**) first, before continuing with the shutdown process.
```sh
...

close() {
  kill -s SIGINT $(pgrep -f "node /usr/local/src/server.js")
  sleep 3
  kill -TERM "$child" 2>/dev/null
}

trap close SIGTERM

npm start &

child=$!
wait "$child"
```

## server.js
It was needed to move the `require` part outside of the micro init function.
```js

let userFunction = require(USER_CODE_PATH + '/' + process.env.INTERNAL_RUNTIME_ENTRYPOINT);

...
```
The reason for that is that the `process.on('SIGINT')` needed to be loaded the second the `pm2` is running the file. And, when the `require` function is **inside** the function the the `process.on` won't be able to listen to the `SIGINT` event. 

As of that, I've also removed the following line in all the micro functions.
```js
let userFunction = require(USER_CODE_PATH + '/' + process.env.INTERNAL_RUNTIME_ENTRYPOINT);
```

### Tests:

- [x] Checked the regular test of `x-internal-challenge`
- [x] Checked the SIGINT with the following code 👇.
I've added this line at the end of each of the `index.js` files inside the `example` folder.
```js
const https = require('https');

process.on('SIGINT', function () {
    https.get(`https://api.telegram.org/bot[TELEGRAM_BOT]/sendMessage?chat_id=[TELEGRAM_CHAT_ID]&text=end`);
})
```

The idea came thanks to Appwrite community member's contributions in [this](https://discord.com/channels/564160730845151244/1117792744178450442/1117792744178450442) Discord thread. 

Helps to (in the Node part): https://github.com/appwrite/appwrite/issues/5688

### DIFF
--- a/runtimes/node-14.5/server.js
+++ b/runtimes/node-14.5/server.js
@@ -4,6 +4,8 @@ const { json, send } = require("micro");
 
 const USER_CODE_PATH = '/usr/code-start';
 
+let userFunction = require(USER_CODE_PATH + '/' + process.env.INTERNAL_RUNTIME_ENTRYPOINT);
+
 const server = micro(async (req, res) => {
     const body = await json(req);
 
@@ -56,7 +58,6 @@ const server = micro(async (req, res) => {
     };
 
     try {
-        let userFunction = require(USER_CODE_PATH + '/' + process.env.INTERNAL_RUNTIME_ENTRYPOINT);
 
         if (!(userFunction || userFunction.constructor || userFunction.call || userFunction.apply)) {
             throw new Error("User function is not valid.")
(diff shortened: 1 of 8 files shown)


nev-21:
Hi @byawitz, please has this been moved to other PR or issue?

byawitz:
it was by mistake when I delete old repositories from my account.
I've recreated it #175

nev-21:
thank you

PR #188, opened 2023-07-28
Github Actions CI/CD

faisalill:


faisalill:
@christyjacob4

ga_deploy.sh
```diff
@@ -0,0 +1,8 @@
+echo 'Starting deployment...'
+
```
christyjacob4, line 2: Is this file and `ga_tests.sh` required outside the context of github actions ? If not, then lets move these sh files to the workflows folder itself

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
...
+    name: ${{ matrix.runtimes }}
+    steps: 
+        - name: Clone Repository
+          uses: actions/checkout@v3.5.3
+        - name: Setup Docker Buildx
+          uses: docker/setup-buildx-action@v2.9.1
+        - name: Check Docker Buildx
+          run: docker buildx version
+        - name: Install PHP
+          uses: shivammathur/setup-php@2.25.4
```
christyjacob4, line 385: Can we look for a more official one ? If not, possibly the most well maintained one?

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
+name: GitHub Actions Demo
```
christyjacob4, line 1:
```suggestion
name: Open Runtimes Tests
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
...
+run-name: OPEN RUNTIMES TEST
```
christyjacob4, line 2:
```suggestion
```
I dont think we need a separate run name

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
...
+on: [push]
+jobs:
+  OPEN_RUNTIMES:
+    strategy:
+        fail-fast: false
+        matrix:
+          runtimes: [node-20.4.js, node-20.4.mjs, node-19.0.js, node-19.0.mjs, node-18.0.js, node-18.0.mjs, node-16.0.js, node-16.0.mjs, node-14.5.js, node-14.5.mjs, deno-1.35, deno-1.24, deno-1.21, python-3.11, python-3.10, python-3.9, python-3.8, dart-3.0, dart-2.19, dart-2.18, dart-2.17, dart-2.16, dart-2.15, ruby-3.2, ruby-3.1, ruby-3.0, php-8.2, php-8.1, php-8.0, swift-5.8, swift-5.5, kotlin-1.6, kotlin-1.8, java-18.0, java-17.0, java-11.0, java-8.0, cpp-17, cpp-20, dotnet-7.0, dotnet-6.0]
+          include:
+            - runtimes: node-20.4.js
```
christyjacob4, line 11: Lets add some comments to give some visual separation between the different environments

Please refer to the current travis file

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
...
+          run : |
+           RUNTIME="${{matrix.RUNTIME}}" ENTRYPOINT="${{ matrix.ENTRYPOINT }}" TEST_CLASS="${{ matrix.TEST_CLASS }}" OPEN_RUNTIMES_BUILD_COMMAND="${{ matrix.OPEN_RUNTIMES_BUILD_COMMAND }}" OPEN_RUNTIMES_START_COMMAND="${{ matrix.OPEN_RUNTIMES_START_COMMAND }}" sh ./tests.sh
+        - name: Login to DockerHub
+          uses: docker/login-action@v2.2.0
+          with:
+            username: ${{ secrets.DOCKER_USERNAME }}
+            password: ${{ secrets.DOCKER_PASSWORD }}
+        - name: Deploy to DockerHub
+          run: |
+           ARCH="${{ matrix.ARCH }}" IMAGE="${{ matrix.IMAGE }}" RUNTIME="${{ matrix.RUNTIME }}" sh ./deploy.sh
```
christyjacob4, line 402: Make sure the deploy step only runs when we create a tag. And the tests run on every push.

Checkout the travis file to see how we handle the deploy stage
https://github.com/open-runtimes/open-runtimes/blob/main/.travis.yml#L411-L412

.github/workflows/tests.yml
faisalill: @christyjacob4 there is no official setup-php action. 
This is was the best option I found.

If this is not needed how about I write a php and composer  installation command ?? rather than using an action.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,404 @@
+name: GitHub Actions Demo
+run-name: OPEN RUNTIMES TEST
+on: [push]
```
christyjacob4, line 3: We need to run on pull_requests on tag creation. please check the right flags for it

christyjacob4 (changes requested):
Awesome work 👏  please address the comments

.github/workflows/tests.yml
Meldiron: On push should cover pull requests. Before making pull request, you must push into a branch.

Yes, I agree we need something for when tag is released (to deploy)

.github/workflows/tests.yml
Meldiron: If possible, let's make tests run on deploy too. Tests first, then deploy. So when tests fail for some reason at that stage, it wont deploy.

faisalill:
@christyjacob4 @Meldiron 
### Testing Plan for asked changes: 
1. Run tests only on push: https://github.com/faisalill/open-runtimes/actions/runs/5726887707/job/15518210087
2. Run tests only on PR: https://github.com/faisalill/open-runtimes/pull/1
3. Run tests before deployment only on tags: https://github.com/faisalill/open-runtimes/actions/runs/5727245693

.github/workflows/tests.yml
faisalill: @Meldiron @christyjacob4 I have mentioned to requested changes below. What do you guys think of it?

.github/workflows/tests.yml
faisalill: @Meldiron whenever a new release with a tag(version) is released all the steps run. It starts with the tests first if they fail it will suspend all the other steps and never execute the deployment step at all.

.github/workflows/tests.yml
abnegate: Let's use PHP inside docker instead if we can :+1:

faisalill: @abnegate That would require the same php and composer install command in every dockerfile of all the runtimes.(if i am correct)
How about I justt use a php and composer install command in a step rather than doing it for every dockerfile.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,454 @@
+name: Open Runtimes Tests
+on: 
+  push:
+    branches:
+      - '*'
+  pull_request:
+    branches:
+      - '*'
+  release: 
+    types: [created]
```
christyjacob4, line 10: these can be combined and made more concise

.github/workflows/tests.yml
christyjacob4: @faisalill alright. We can keep using it then

christyjacob4 (changes requested):
Please address the comments and then we're good to merge

.github/workflows/tests.yml
```diff
@@ -0,0 +1,448 @@
...
+          run: composer install
+        - name: Run Tests
+          run : |
+           RUNTIME="${{matrix.RUNTIME}}" ENTRYPOINT="${{ matrix.ENTRYPOINT }}" TEST_CLASS="${{ matrix.TEST_CLASS }}" OPEN_RUNTIMES_BUILD_COMMAND="${{ matrix.OPEN_RUNTIMES_BUILD_COMMAND }}" OPEN_RUNTIMES_START_COMMAND="${{ matrix.OPEN_RUNTIMES_START_COMMAND }}" sh ./tests.sh
+        - if: startsWith(github.ref, 'refs/tags/')
+          name: Login to DockerHub
+          uses: docker/login-action@v2.2.0
+          with:
+            username: ${{ secrets.DOCKER_USERNAME }}
+            password: ${{ secrets.DOCKER_PASSWORD }}
```
faisalill, line 442: @christyjacob4 
Secrets to be set in github secrets are these. 
Should it be `DOCKER_USERNAME` and `DOCKER_PASSWORD` or something different.

.github/workflows/tests.yml
faisalill: Done

.github/workflows/tests.yml
```diff
@@ -0,0 +1,448 @@
...
+
+on: [push, pull_request, release]
+
+jobs:
+  OPEN_RUNTIMES:
+    strategy:
+        fail-fast: false
+        matrix:
+          runtimes: [node-20.0.js, node-20.0.mjs, node-19.0.js, node-19.0.mjs, node-18.0.js, node-18.0.mjs, node-16.0.js, node-16.0.mjs, node-14.5.js, node-14.5.mjs, deno-1.35, deno-1.24, deno-1.21, python-3.11, python-3.10, python-3.9, python-3.8, dart-3.0, dart-2.19, dart-2.18, dart-2.17, dart-2.16, dart-2.15, ruby-3.2, ruby-3.1, ruby-3.0, php-8.2, php-8.1, php-8.0, swift-5.8, swift-5.5, kotlin-1.6, kotlin-1.8, java-18.0, java-17.0, java-11.0, java-8.0, cpp-17, cpp-20, dotnet-7.0, dotnet-6.0]
+          include:
```
stnguyen90, line 11: Do we need the runtimes or can we just use include? I read this: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations.

Otherwise, it would be good to try and figure out a combination that would cut down on the includes array a bit. Like maybe runtimes would be just the language and then include adds/expands to include different versions? Maybe not..I'm not really sure, but it would be nice if we can reduce some of the repetition in here.

faisalill: @stnguyen90 
> Do we need the runtimes or can we just use include? I read this: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-adding-configurations.

Changed from using runtimes array to just include. 

> Otherwise, it would be good to try and figure out a combination that would cut down on the includes array a bit. Like maybe runtimes would be just the language and then include adds/expands to include different versions? Maybe not..I'm not really sure, but it would be nice if we can reduce some of the repetition in here.

How about having different test files like node.yml, deno.yml for each runtime as each version of node has to run twice (for tests.js and tests.mjs), few runtimes exclude ceratain architecture, we can't have an array of language and another for versions conditionally coz arrays inside -include don't work.(below mentioned code) [My Attempt](https://github.com/faisalill/open-runtimes/actions/runs/5796604844/job/15710511738)
```yml
...
matrix: 
    languages: [node,deno,python...]
    include: 
        - languages: node
          versions: [20.0, 19.0, 18.0...]
```

So how about having different .yml files for each language??

.github/workflows/tests.yml
```diff
@@ -0,0 +1,415 @@
+name: Open Runtimes Tests
+
+on: [push, pull_request, release]
```
christyjacob4, line 3:
```suggestion
on: [pull_request]
```

faisalill: @christyjacob4 committed i

loks0n:
Closing as duplicate

PR #190, opened 2023-08-21
feat: set utf-8 default

loks0n:
Sets charset to utf-8 for every runtime given:
- charset is not already set
- the content type is not a multipart type

### DIFF
--- a/runtimes/cpp-20/src/RuntimeResponse.h
+++ b/runtimes/cpp-20/src/RuntimeResponse.h
@@ -10,18 +10,24 @@ namespace runtime
     class RuntimeResponse
     {
         public:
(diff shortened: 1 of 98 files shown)


runtimes/cpp-20/src/RuntimeResponse.h
```diff
@@ -1,6 +1,7 @@
 #ifndef CPP_RUNTIME_RUNTIMERESPONSE_H
 #define CPP_RUNTIME_RUNTIMERESPONSE_H
 
+#include <fstream>
```
Meldiron, line 4: Feels random, is this nessessary?

Meldiron (approved):

PR #191, opened 2023-08-25
feat: show native logs

loks0n:
Show native logs inside a disclaimer, instead of hiding them

Swift and Go have been left out


Meldiron (approved):
Amazing quality 🔥

PR #195, opened 2023-09-11
Feat: Bun 1.0 runtime

Meldiron:
This adds Bun runtime into Open Runtimes. Later it can be used in Appwrite.

Tasks:

- [x] Create runtime from Deno (take from Node what is needed)
- [x] Use Deno HTTP server
- [x] Implement tests

Related issue:

* https://github.com/appwrite/appwrite/issues/3538

### DIFF
--- a/tests.sh
+++ b/tests.sh
@@ -2,7 +2,7 @@
 
 # Build runtime Docker image
 cd runtimes/${RUNTIME}
-docker build -t open-runtimes/test-runtime .
+docker build --load -t open-runtimes/test-runtime .
 
 # Enter function folder
 cd ../../


runtimes/bun-1.0/helpers/build.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+# Fail build if any command fails
+set -e
+
+. /usr/local/server/helpers/before-build.sh
```
christyjacob4, line 5: what is this `.` supposed to be doing?

christyjacob4 (approved):
please check the comment

runtimes/bun-1.0/helpers/build.sh
stnguyen90: it's the same as [`source`](https://linuxize.com/post/bash-source-command/) which is kind of like an import and execute the script in the current context.

Meldiron: Yep, it's replacement for source. We use it in ~half of runtimes. Some (for some reason) simply don't have source. `.` is drop-in replacement, seems like. So far I didn't spot any issue with it

Meldiron: For example in Java: https://github.com/open-runtimes/open-runtimes/blob/main/runtimes/java-18.0/helpers/build.sh

christyjacob4: Alright got it 👍

PR #197, opened 2023-09-25
Feat: Dart 3.1

Meldiron:
Adds Dart 3.1 support, released on August 16


.travis.yml
```diff
@@ -158,6 +158,14 @@ env:
       OPEN_RUNTIMES_START_COMMAND="python3 src/server.py"
 
     # Dart
+    - RUNTIME=dart-3.1
+      TEST_CLASS=Base
+      ENTRYPOINT=lib/tests.dart
+      IMAGE=openruntimes/dart:${VERSION}-3.0
```
lohanidamodar, line 164:
```suggestion
      IMAGE=openruntimes/dart:${VERSION}-3.1
```

lohanidamodar (changes requested):
Small typo, rest looks good.

.travis.yml
Meldiron: Fixed, good catch

lohanidamodar (approved):

runtimes/dart-3.1/example/pubspec.yaml
```diff
@@ -0,0 +1,12 @@
+name: open_runtimes_example
+description: Dart runtime example
+version: 1.0.0
+publish_to: none
+# homepage: https://www.example.com
+
+environment:
+  sdk: '>=3.0.0 <4.0.0'
```
christyjacob4, line 8: Does this need to be 3.1 ? @Meldiron

Meldiron: Not too familiar with Dart release cycle, but this is what we do for 2.x versions: https://github.com/open-runtimes/open-runtimes/blob/main/runtimes/dart-2.15/pubspec.yaml#L8

We always od 2.12 - 3.0. I am guessing 2.12 had breaking change.

So until we see some breaking changes inside 3.x, we should be doing 3.0 to 4.0.

Meldiron: If we limited this to exactly 3.0, then build will probably fail when developer uses some library that is for 3.1.
Which would be very annoying, and we would have to have every single version on Cloud to ensure all versions can be used with Appwrite.

I would rather address the problem when it pops up, disallowing versions that has breaking changes.

christyjacob4 (approved):

PR #205, opened 2023-11-13
fix: Properly serialize native errors passed to ctx.error()

### LINKED ISSUE #204: Native JS errors don't properly serialize
Currently, JSON.stringify() is being called on the object passed to ctx.error() if it's an Object - which sadly does not work on Error classes, as they serialize to `{}`.
This results in `ctx.error(new Error('some message'))` outputting `{}`, which is sure to confuse people, and I'm guessing isn't intended behaviour.

A minimal patch would be to check if an object is an instanceof Error, and if so, return `err.stack ?? err.message`.

Exponential-Workload:
When encountering an Error class passed to ctx.error(), this PR uses error.stack, falling back to error.message, to have a proper error message.
The issue with just calling JSON.stringify() is that the built-in JS Error class stringifies to `{}`.

This PR fixes #204.

### DIFF
--- a/runtimes/node-20.0/src/server.js
+++ b/runtimes/node-20.0/src/server.js
@@ -96,7 +96,9 @@ const server = micro(async (req, res) => {
             }
         },
         error: function (message) {
-            if (message instanceof Object || Array.isArray(message)) {
+            if (err instanceof Error) {
+                errors.push(err.stack ?? err.message);
+            } else if (message instanceof Object || Array.isArray(message)) {
                 errors.push(JSON.stringify(message));
             } else {
                 errors.push(message + "");


runtimes/node-20.0/src/server.js
```diff
@@ -96,7 +96,9 @@ const server = micro(async (req, res) => {
             }
         },
         error: function (message) {
-            if (message instanceof Object || Array.isArray(message)) {
+            if (err instanceof Error) {
+                errors.push(err.stack ?? err.message);
```
AlphanB, line 100: It has to be `message` instead of `err`. Thank you.

Exponential-Workload:
Might make more sense to just overall, for everything, use [util.format()](https://nodejs.org/api/util.html#utilformatformat-args) - which is what the `console` library uses. Will adjust code to match if this is desired.

loks0n:
> Might make more sense to just overall, for everything, use [util.format()](https://nodejs.org/api/util.html#utilformatformat-args) - which is what the `console` library uses. Will adjust code to match if this is desired.

If we can achieve this without breaking changes that would be ideal

PR #213, opened 2024-01-02
Feat: Internal error logging

Meldiron:
Description:

Open Runtime code could fail depending on the input provided. This PR adds middleware that ensures such errors are not in the response body (default behaviour in many HTTP servers), instead, send it in errors header (stored more securely).

Example of such error is empty body when content type header says JSON.

Related PRs:

- https://github.com/appwrite/appwrite/issues/6431


lohanidamodar (commented):
I can't see to understand what's changed (looking at changes in Dart runtimes)

eldadfux (approved):

PR #216, opened 2024-02-15
Fix: Improve build speed

Meldiron:
If a code change is made, permissions on helpers is set each time. For some reason, this is slow and takes 3-5 seconds (on Gitpod workspace)

Doing this trick in Dockerfile ensures that helper setup is only made when there is a change to helper files.

Before:

https://github.com/open-runtimes/open-runtimes/assets/19310830/b659effb-c62b-4ce9-bbd8-0535c3e1a973

After:

https://github.com/open-runtimes/open-runtimes/assets/19310830/4d4fba3d-7978-4321-9cc6-4c05d21b4564

### DIFF
--- a/runtimes/bun-1.0/Dockerfile
+++ b/runtimes/bun-1.0/Dockerfile
@@ -16,7 +16,7 @@ COPY package* /usr/local/server/
 
 RUN bun install
 
-COPY . .
+COPY ./helpers ./helpers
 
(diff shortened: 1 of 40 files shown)


christyjacob4 (approved):

Meldiron:
CleanShot 2024-02-18 at 10 03 49@2x

Dont merge. It might be causing this.

PR #217, opened 2024-02-15
Feat: Go 1.22

Meldiron:
After learning Golang, I implemented Go runtime with concurrency support, better structure and better dependencies management compared to previous Go 1.19 implementation.

It also implements Open Runtimes V3 structure, so it's ready to be used with Appwrite 🎉

This is an initial step towards Golang in Appwrite Cloud.


tests/resources/functions/go-1.22/tests.go
```diff
@@ -0,0 +1,156 @@
...
+	"openruntimes/types"
+	"os"
+)
+
+func Main(Context *types.Context) types.ResponseOutput {
+	action := Context.Req.Headers["x-action"]
+
+	switch a := action; a {
+	case "plaintextResponse":
+		return Context.Res.Send("Hello World 👋", 200, nil)
```
Meldiron, line 13: Looking for review, is this typical in Golang?

I was considering `context.res.Send()` instead, but then I saw Go code example `os.Stdout = writer` which made me think variables are too first-uppercase. Not 100% confident.

runtimes/go-1.22/helpers/after-build.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/sh
+# Fail build if any command fails
+set -e
+
+echo "Compiling ..."
+
+# Compile the Code
+cd /usr/local/server/src
+go get openruntimes/handler@v0.0.0
+go build
```
TorstenDittmann, line 10:
```suggestion
go build -ldflags="-s -w"
```
this will build for production stripping off any off information not needed during execution

runtimes/go-1.22/src/main.go
```diff
@@ -0,0 +1,307 @@
...
+		}
+
+		if len(queryChunk) > 1 {
+			queryValue = queryChunk[1]
+		}
+
+		if queryKey != "" {
+			query[queryKey] = queryValue
+		}
+	}
```
TorstenDittmann, line 140: you can use https://pkg.go.dev/net/url#ParseQuery 👍🏻

Meldiron: Applied

Meldiron (commented):
LGTM

runtimes/go-1.22/src/main.go
```diff
@@ -0,0 +1,260 @@
...
+			Host:        host,
+			Port:        port,
+			Path:        path,
+			QueryString: queryString,
+			Query:       query,
+		},
+		Res: types.Response{},
+	}
+
+	context.Req.SetBodyBinary(bodyBytes)
```
byawitz, line 140: Why is `bodyBytes` not in the constructor?

byawitz (approved):

PR #240, opened 2024-05-13
Fix: Binary support

Meldiron:
Currently, Node parses all bodies as texts. This PR adds support for buffer too. Also, it addresses same problem in other rubtimes.

Implemented in backwards-compatible approach.


Meldiron:
Closed in favour of https://github.com/open-runtimes/open-runtimes/pull/251

PR #242, opened 2024-05-25
Add `HEAD` & `TRACE` Support

ItzNotABug:
The runtimes use an external library to manage the http server where not all the underlying libraries/modules support `TRACE` method. Below is a list of the one's I've tested with findings -

---
### Tested
* Dart-3.3
* Bun-1.0
* Ruby-3.3
* Kotlin 1.9
* Deno-1.40
* Node-21.0
* Python-3.12
* Python-ML-3.11
* Swift-5.9, 5.8, 5.5

---
### `TRACE` not supported -
* PHP
* C++ - 17, 20 (Drogon Library)
   There might be a possibility of using a Middleware approach but I am not entirely sure on that

### DIFF
--- a/runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 import io.javalin.Javalin
 import io.javalin.http.Context
+import io.javalin.http.HandlerType
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeoutOrNull
 import java.io.*
@@ -28,7 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
-        .trace("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -28,6 +28,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .trace("/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/kotlin-1.8/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.8/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 import io.javalin.Javalin
 import io.javalin.http.Context
+import io.javalin.http.HandlerType
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeoutOrNull
 import java.io.*
@@ -28,7 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
-        .trace("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/kotlin-1.8/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.8/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -28,6 +28,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .trace("/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/kotlin-1.9/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.9/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -4,6 +4,7 @@ import com.google.gson.Gson
 import com.google.gson.GsonBuilder
 import io.javalin.Javalin
 import io.javalin.http.Context
+import io.javalin.http.HandlerType
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withTimeoutOrNull
 import java.io.*
@@ -28,7 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
-        .trace("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/kotlin-1.9/src/main/kotlin/io/openruntimes/kotlin/Server.kt
+++ b/runtimes/kotlin-1.9/src/main/kotlin/io/openruntimes/kotlin/Server.kt
@@ -28,6 +28,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .trace("/*") { runBlocking { execute(it) } }
 }
 
 suspend fun execute(ctx: Context) {
--- a/runtimes/python-3.10/src/server.py
+++ b/runtimes/python-3.10/src/server.py
@@ -69,7 +69,7 @@ class Context:
         else:
             self.errors.append(str(message))
 
-HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
+HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'PATCH']
 
 async def action(request):
     timeout = request.headers.get('x-open-runtimes-timeout', '')
--- a/runtimes/ruby-3.3/src/server.rb
+++ b/runtimes/ruby-3.3/src/server.rb
@@ -4,6 +4,15 @@ require 'async'
 
 USER_CODE_PATH = '/usr/local/server/src/function';
 
+configure do
+  class << Sinatra::Base
+    def trace(path, opts={}, &block)
+      route 'TRACE', path, opts, &block
+    end
+  end
+  Sinatra::Delegator.delegate :trace
+end
+
 class RuntimeResponse
   def send(body, status_code = 200, headers = {})
     {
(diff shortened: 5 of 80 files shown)


tests/Base.php
```diff
@@ -213,6 +213,15 @@ public function testRequestMethod(): void
         $response = $this->execute(method: 'PATCH', headers: ['x-action' => 'requestMethod']);
         self::assertEquals(200, $response['code']);
         self::assertEquals('PATCH', $response['body']);
+
+        $response = $this->execute(method: 'HEAD', headers: ['x-action' => 'requestMethod']);
+        self::assertEquals(200, $response['code']);
+        // Some runtimes either return '', while some return 'HEAD'
+        // self::assertEquals('', $response['body']);
```
Meldiron, line 220: Let's enable this test and allow both. Refference to allowing 2 options in a test: https://github.com/ItzNotABug/open-runtimes/blob/684549e12f49def843cc2162428ab2f7c5588f33/tests/Base.php#L241

tests/Base.php
```diff
@@ -213,6 +213,15 @@ public function testRequestMethod(): void
         $response = $this->execute(method: 'PATCH', headers: ['x-action' => 'requestMethod']);
```
Meldiron, line 213: Above this test is `OPTIONS`, seems to only be problematic on C++. Let's please enable it as part of this PR. Ill also take a look at all failing tests, and give helping hand. If neither of us can figure it out, we can again disable the test, but specifically for C++ only.

tests/Base.php
ItzNotABug: Makes sense. I'll enable this one.

tests/Base.php
ItzNotABug: Sure thing.

ItzNotABug:
I guess the Kotlin tests failed too, I'll double check the `javalin`'s impl.

tests/Base.php
```diff
@@ -209,10 +209,20 @@ public function testRequestMethod(): void
...
+        self::assertContains($response['body'], ['OPTIONS', '']);
 
         $response = $this->execute(method: 'PATCH', headers: ['x-action' => 'requestMethod']);
         self::assertEquals(200, $response['code']);
         self::assertEquals('PATCH', $response['body']);
+
+        $response = $this->execute(method: 'HEAD', headers: ['x-action' => 'requestMethod']);
+        self::assertEquals(200, $response['code']);
+        //  var_dump says the body is a boolean in C++.
+        self::assertContains($response['body'], ['HEAD', '', false]);
```
ItzNotABug, line 221: @Meldiron,
One thing to note here is that if I do below, the body is properly an empty string -
```php
self::assertContains(trim($response['body']), ['HEAD', '']);
```
else its a boolean, atleast as per the `var_dump`.

byawitz:
@ItzNotABug the Kotlin would is a x64 image, to run it on an arm machines as Apple `mX` processors run it with the `--platform linux/amd64` flag 

image

ItzNotABug:
@byawitz Thanks for the tip!
It worked & I was able to run the tests!

tests/Base.php
ItzNotABug: Something similar happens on the Kotlin side as well -
```php
["body"]=> bool(false)
```

tests/Base.php
```diff
@@ -10,8 +10,11 @@
 
 class Base extends TestCase
 {
+    private string $runtime = '';
+
     public function setUp(): void
     {
+        $this->runtime = \explode('-', \getenv('RUNTIME'), 2)[0] ?? '';
```
Meldiron, line 17: This wont work well for `python-ml-3.11`, for example. We need to split by `-`, remove last from array, and join back with `-`. With that logic we get `python-ml` proplerly.

tests/Base.php
```diff
@@ -207,14 +212,40 @@ public function testRequestMethod(): void
...
+
+        $response = $this->execute(method: 'TRACE', headers: ['x-action' => 'requestHeadOrTrace']);
+        if($this->runtime === 'php') {
+            // Bug in PHP Swoole
+            self::assertEquals(400, $response['code']);
+        } else {
+            self::assertEquals(200, $response['code']);
+        }
+
+        self::assertStringContainsString('TRACE', $response['headers']['x-open-runtimes-logs']);
```
Meldiron, line 246: Based on: https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/TRACE

Seems like TRACE should be allowed to have body. What's the reason to use logs here?

tests.sh
```diff
@@ -1,13 +1,40 @@
...
+
+if ! command -v docker &> /dev/null; then
+    echo "Docker is not installed or running!"
+    exit 1
+fi
+
+if ! command -v timeout &> /dev/null; then
+    echo "timeout is not available."
+    exit 1
+fi
```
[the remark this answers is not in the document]
ItzNotABug: I think the `timout` isn't directly available on a Mac system. Needs to be added as a separate lib from `brew`. Or did I miss something?

tests/Base.php
ItzNotABug: If I remember correctly, we discussed using the logs for both `Head` & `TRACE` as one of the runtimes did not return a body for `TRACE` as well. I will find the exact convo. and ping it again to you.

tests/Base.php
ItzNotABug: Makes sense. Will check.

tests/Base.php
```diff
@@ -207,14 +212,40 @@ public function testRequestMethod(): void
...
         self::assertEquals('PATCH', $response['body']);
     }
 
+    public function testHeadAndTraceRequestMethods(): void
+    {
+        $response = $this->execute(method: 'HEAD', headers: ['x-action' => 'requestHeadOrTrace']);
+        self::assertEquals(200, $response['code']);
+
+        // head, trace dont work well with cpp [drogon].
+        if($this->runtime === 'cpp') return;
```
Meldiron, line 234: Not sure if I like this, I would rather have 2 larget if/else, so I know what exactly is expected for CPP. Here it seems to be expected to give 200, but not proper logs.

Meldiron (commented):
Last review, good job! Let's address remaining comments and Ill mark as approved

tests/Base.php
ItzNotABug: Only the response-code works for CPP, so for that runtime, I returned instead of moving ahead with other tests as they'd eventually be skipped due to an `if/else` check. But I'll add anyway. Better if we have descriptive logs/comments.

ItzNotABug:
> Last review, good job! Let's address remaining comments and Ill mark as approved

Will check the review comments and the failing php tests too. Thanks.

tests.sh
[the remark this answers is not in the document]
Meldiron: In one of the main PRs for next release, we removed `timeout` logic from this test file: https://github.com/open-runtimes/open-runtimes/blob/feat-log-files/tests.sh

It didn't work well for some scenarios, it was flaky. Doing simple sleep worked better.

tests/Base.php
Meldiron: Thank you so much for taking time to ensure quality of this repository 🙏
Let's switch back to body response, and have `if` to skip the test for the runtime that doesn't support it.

Later we can fix those separately, but using proper body test ensures quality for the rest of runtimes.

tests/Base.php
Meldiron: Understandable, I would prefer this structrure of the test then:

```
if runtiem is cpp
	call()
	ensure 200()
else
	call()
	ensure 200()
	more tests...
```

Its a bit of code duplication but it's for edge-case that isn't supposed to be there, so it's good it stands out.

tests.sh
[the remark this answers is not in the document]
ItzNotABug: I see, thanks. I'll move to using `sleep`.

tests/Base.php
ItzNotABug: Sure thing.

tests/Base.php
ItzNotABug: Agree. Will address this too.

tests.sh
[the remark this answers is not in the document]
Meldiron: No need, for now we can just avoid adding this check for timeout command being available. Later when that other PR is merged (very soon), the sleep will be in use.

tests/Base.php
ItzNotABug: Done.

tests/Base.php
ItzNotABug: As discussed, most likely the web servers wouldn't really return anything. Hence the response body check is not added.

tests/Base.php
ItzNotABug: Done.

tests.sh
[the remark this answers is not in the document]
ItzNotABug: Done.

Meldiron (approved):

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
```diff
@@ -28,6 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
```
christyjacob4, line 32: as per the docs, https://javalin.io/documentation#handlers you can use `trace()` directly.

runtimes/kotlin-1.8/src/main/kotlin/io/openruntimes/kotlin/Server.kt
```diff
@@ -28,6 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
```
christyjacob4, line 32: as per the docs, https://javalin.io/documentation#handlers you can use `trace()` directly.

runtimes/kotlin-1.9/src/main/kotlin/io/openruntimes/kotlin/Server.kt
```diff
@@ -28,6 +29,7 @@ suspend fun main() {
         .patch("/*") { runBlocking { execute(it) } }
         .options("/*") { runBlocking { execute(it) } }
         .head("/*") { runBlocking { execute(it) } }
+        .addHandler(HandlerType.TRACE, "/*") { runBlocking { execute(it) } }
```
christyjacob4, line 32: as per the docs, https://javalin.io/documentation#handlers you can use `trace()` directly.

runtimes/python-3.10/src/server.py
```diff
@@ -69,7 +69,7 @@ def error(self, message):
         else:
             self.errors.append(str(message))
 
-HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'CONNECT', 'OPTIONS', 'TRACE', 'PATCH']
+HTTP_METHODS = ['GET', 'HEAD', 'POST', 'PUT', 'DELETE', 'OPTIONS', 'TRACE', 'PATCH']
```
christyjacob4, line 72: why are we removing this ?

runtimes/ruby-3.3/src/server.rb
```diff
@@ -310,6 +319,10 @@ def handle(request, response)
   handle(request, response)
 end
 
+trace '*' do
+  handle(request, response)
+end
+
```
christyjacob4, line 325: what about head requests for ruby? I dont see a handler for it?

christyjacob4 (changes requested):
Awesome Work!! 👏  Just a couple of questions

runtimes/kotlin-1.6/src/main/kotlin/io/openruntimes/kotlin/Server.kt
ItzNotABug: Docs mention supporting `TRACE`, however the routing api doesn't have a dedicated `trace()` method. I actually tried it earlier, the tests failed as there was no method. Ref - https://github.com/open-runtimes/open-runtimes/pull/242/commits/5f6085239ea45cbca691eca2fd6bf56cd80658d5

runtimes/python-3.10/src/server.py
ItzNotABug: If I recall correctly, it was to keep consistent methods across runtimes. Not all of them support `CONNECT`.

runtimes/ruby-3.3/src/server.rb
ItzNotABug: The `get` handler auto-registers one for `head` too, [here's the reference](https://github.com/sinatra/sinatra/blob/main/lib/sinatra/base.rb#L1524C7-L1532C10).

Its been a while since I tested ruby, but the `head` handler's request are probably overriden with the `get`.

PR #243, opened 2024-05-30
Feat: Labels and ENV secret

Meldiron:
- Add label to identify OPR images
- Add default to secret ENV var
- Allow no secret header if secret ENV is empty

### DIFF
--- a/runtimes/cpp-17/src/main.cc
+++ b/runtimes/cpp-17/src/main.cc
@@ -56,7 +56,7 @@ int main()
                     }
 
                     std::string secret = req->getHeader("x-open-runtimes-secret");
-                    if (secret.empty() || secret != std::getenv("OPEN_RUNTIMES_SECRET"))
+                    if (!std::getenv("OPEN_RUNTIMES_SECRET").empty() && secret != std::getenv("OPEN_RUNTIMES_SECRET"))
                     {
                         res->setStatusCode(static_cast<drogon::HttpStatusCode>(500));
                         res->setBody("Unauthorized. Provide correct \"x-open-runtimes-secret\" header.");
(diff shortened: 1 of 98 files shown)


runtimes/cpp-20/src/main.cc
```diff
@@ -56,7 +56,7 @@ int main()
                     }
 
                     std::string secret = req->getHeader("x-open-runtimes-secret");
-                    if (secret.empty() || secret != std::getenv("OPEN_RUNTIMES_SECRET"))
+                    if (std::getenv("OPEN_RUNTIMES_SECRET") != nullptr && secret != std::getenv("OPEN_RUNTIMES_SECRET"))
                     {
                         res->setStatusCode(static_cast<drogon::HttpStatusCode>(500));
```
christyjacob4, line 61: the dockerfile has not been updated?

runtimes/cpp-17/src/main.cc
```diff
@@ -56,7 +56,7 @@ int main()
                     }
 
                     std::string secret = req->getHeader("x-open-runtimes-secret");
-                    if (secret.empty() || secret != std::getenv("OPEN_RUNTIMES_SECRET"))
+                    if (std::getenv("OPEN_RUNTIMES_SECRET") != nullptr && secret != std::getenv("OPEN_RUNTIMES_SECRET"))
```
christyjacob4, line 59: the dockerfile has not been updated?

runtimes/dotnet-6.0/src/Program.cs
```diff
@@ -51,7 +51,9 @@ static async Task<IResult> Action(HttpRequest request)
         ? secretValue.ToString()
         : string.Empty;
 
-    if(secret == string.Empty || secret != (Environment.GetEnvironmentVariable("OPEN_RUNTIMES_SECRET") ?? ""))
+
```
christyjacob4, line 54: dockerfile not updated ??

christyjacob4 (approved):

PR #245, opened 2024-05-31
Feat: Log files

Meldiron:
Use log files instead of headers

- [x] Update existing tests to look at log files
- [x] Add tests for new headers


runtimes/node-18.0/src/server.js
```diff
@@ -5,20 +5,40 @@ const { text: parseText, json: parseJson, send } = require("micro");
 
 const USER_CODE_PATH = '/usr/local/server/src/function';
 
+const randomId = () => {
+    return `${new Date().getTime().toString(16)}${Math.round(Math.random() * 1000000000).toString(16)}`;
+}
+
+const writeLogs = (id, type, log) => {
+    fs.appendFileSync(`/mnt/logs/${id}_${type}.log`, log);
```
christyjacob4, line 13: instead of writing to a file like this why not use a more native approach 

```
var logsStdout = fs.createWriteStream('mnt/logs/${id}_stdout.log');
var logsStderr = fs.createWriteStream('mnt/logs/${id}_stderr.log');
process.stdout.write = logsStdout.write.bind(logsStdout);
process.stderr.write = logsStderr.write.bind(logsStderr);
```

runtimes/node-18.0/src/server.js
```diff
@@ -5,20 +5,40 @@ const { text: parseText, json: parseJson, send } = require("micro");
...
 const server = micro(async (req, res) => {
     try {
         await action(req, res);
     } catch(e) {
-        const logs = [];
-        const errors = [e.stack || e];
-        res.setHeader('x-open-runtimes-logs', encodeURIComponent(logs.join('\n')));
-        res.setHeader('x-open-runtimes-errors', encodeURIComponent(errors.join('\n')));
+        const logging = req.headers[`x-open-runtimes-logging`] ?? 'enabled';
+        const logId = req.headers[`x-open-runtimes-log-id`] ?? randomId();
```
christyjacob4, line 21: Why are we passing an additional logID? we should use the execution ID itself

Meldiron: We dont have execution ID in open runtimes yet. Log ID will be execution ID. Do you want to also call it execution ID? I though tit's best to call id "log-id" because it goes with "logging" header. I thought it would be hard to explain why execution-id is used when logging is enabled

runtimes/node-18.0/src/server.js
Meldiron: Discussed together, we will be keeping approach, but will use streams instead

Meldiron: Applied stream approach

runtimes/node-18.0/src/server.js
```diff
@@ -1,23 +1,26 @@
...
-        await action(req, res);
+        await action(logger, req, res);
     } catch(e) {
-        const logs = [];
-        const errors = [e.stack || e];
-        res.setHeader('x-open-runtimes-logs', encodeURIComponent(logs.join('\n')));
-        res.setHeader('x-open-runtimes-errors', encodeURIComponent(errors.join('\n')));
+        logger.write(e, Logger.TYPE_ERROR);
+        res.setHeader('x-open-runtimes-log-id', logger.id);
+        await logger.end();
```
christyjacob4, line 15: You have the `end()` code in finally as well. so you can remove it from the catch block

runtimes/node-18.0/src/server.js
```diff
@@ -1,23 +1,26 @@
...
+
         return send(res, 500, '');
+    } finally {
+        await logger.end();
```
christyjacob4, line 19: since finally executes in the end, you can also remove the logger.end() in Line 205

byawitz (approved):

PR #248, opened 2024-06-05
Feat: Enforced headers

Meldiron:
Ability to set headers on very request. Useful for local development of functions

- [x] Add test


christyjacob4 (approved):

runtimes/bun-1.0/src/server.ts
```diff
@@ -38,6 +38,11 @@ const action = async (logger: Logger, request: any) => {
     headers[header.toLowerCase()] = request.headers.get(header);
   });
 
+  const enforcedHeaders = JSON.parse(Bun.env["OPEN_RUNTIMES_HEADERS"] ? Bun.env["OPEN_RUNTIMES_HEADERS"] : '{}');
+  for(const header in enforcedHeaders) {
```
christyjacob4, line 42:
```suggestion
  for (const header in enforcedHeaders) {
```
can we run some linter to fix these ?

runtimes/deno-1.24/src/server.ts
```diff
@@ -58,6 +58,11 @@ const action = async (logger: Logger, ctx: any) => {
     headers[header.toLowerCase()] = ctx.request.headers.get(header);
   });
 
+  const enforcedHeaders = JSON.parse(Deno.env.get("OPEN_RUNTIMES_HEADERS") ? (Deno.env.get("OPEN_RUNTIMES_HEADERS") ?? '{}') : '{}');
```
christyjacob4, line 61:
```suggestion
  const enforcedHeaders = JSON.parse(Deno.env.get("OPEN_RUNTIMES_HEADERS") ?? '{}');
```
why not just this?

runtimes/bun-1.0/src/server.ts
```diff
@@ -38,6 +38,11 @@ const action = async (logger: Logger, request: any) => {
     headers[header.toLowerCase()] = request.headers.get(header);
   });
 
+  const enforcedHeaders = JSON.parse(Bun.env["OPEN_RUNTIMES_HEADERS"] ? Bun.env["OPEN_RUNTIMES_HEADERS"] : '{}');
```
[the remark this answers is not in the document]
Meldiron: It doesnt, because if you do `VAR=` (leaving it empty), it will be empty string. Emtry string is a string so `??` wont do anything. Then it goes to JSON.parse, and empty string cannot be parsed - throwing error as result.

runtimes/bun-1.0/src/server.ts
Meldiron: I can make myself task to setup linter (+test) right after v4, right after merging runtime versions together. For now I would like to avoid to keep this PRs simple. Linter never ran here as far as I can remember.

runtimes/deno-1.24/src/server.ts
Meldiron: Same as https://github.com/open-runtimes/open-runtimes/pull/248#discussion_r1638043411 , sadly it wont work for use-case with ENV variables

PR #251, opened 2024-06-09
Feat: Binary request and response

Meldiron:


runtimes/node-21.0/src/server.js
```diff
@@ -72,8 +63,32 @@ const action = async (req, res) => {
...
     const context = {
         req: {
-            bodyRaw,
-            body,
+            get body() {
+                if(contentType.includes("application/json")) {
+                    return this.bodyJson;
+                }
+
+                const binaryTypes = ["audio/", "video/", "octet", "binary"];
```
abnegate, line 71: What about pdfs, archives, vendor specific things like excel/word etc.

Meldiron: I would prefer to have binary as fallback, and specify what is text. But due to backwards compatibility, we have to do it this way. 
Is there an official list of binary content types?

If not, we can make ours and roll with it - `req.body` is supposed to be black-box, so it's fine if it doesn't work 100% as someone might expect. If they know they want binary, they should use `req.bodyBinary`. We keed `req.body` for backwards compatibility, I would personally never suggest it. We should also update all docs to use new `req.bodyText` to indicate it.

abnegate: The more specific we can be the better. The default fallback for binary should be octet-stream, but in that case, the developer will have no idea what the file actually is. If I have a function that handles files, I'm definitely going to want to know if the file is a pdf or a zip without having to check the bytes myself.

Meldiron: You can always use `req.bodyBinary` and look at `req.headers['content-type']` to achievethe same logic. Idea of `req.body` is mainly backwards compatibility, and a bit of magic (for beginners, mostly).

abnegate: If the content-type is `application/pdf`, `req.body` will give me `bodyText`

abnegate: I understand, but the fallback is challenging, if there is _any_ type we don't handle, the function will break (in most languages). There are thousands of possible types: https://www.iana.org/assignments/media-types/media-types.xhtml#application

Meldiron: I will include "application" into check, since most (if not all) of them are binaries

runtimes/dart-2.18/src/function_types.dart
```diff
@@ -14,6 +13,39 @@ class RuntimeRequest {
...
+      if(contentType.startsWith(binaryType)) {
+        return this.bodyBinary;
+      }
+    }
+
+    return this.bodyText;
+  }
+
+  String get bodyRaw {
+    return this.bodyText;
```
christyjacob4, line 38: why does bodyRaw return the parsed text? bodyRaw is supposed to be the unprocessed binary data right ?

Meldiron: `bodyRaw` is not supposed to exist, we only keep it for backwards compatibility. and in V3, `bodyRaw` was string, because we had wrong assumption that bodies are always string.

christyjacob4 (approved):

byawitz (approved):

PR #252, opened 2024-06-09
Feat: Chunks support

Meldiron:


runtimes/node-21.0/src/server.js
```diff
@@ -197,6 +237,12 @@ const action = async (req, res) => {
             if(!executed) {
                 context.error('Execution timed out.');
                 output = context.res.send('', 500, {});
```
Meldiron, line 239: Remove this

tests/Client.php
```diff
@@ -36,6 +32,10 @@ public static function execute($body = '', $url = '/', $method = 'POST', $header
         
                 $key = strtolower(trim($header[0]));
                 $responseHeaders[$key] = trim($header[1]);
+
+                if(\in_array($key, ['x-open-runtimes-logs', 'x-open-runtimes-errors'])) {
```
Meldiron, line 36: TODO: Remove

tests/Base.php
```diff
@@ -462,6 +462,85 @@ public function testDeprecatedMethods(): void
...
+        self::assertEmpty(Client::getErrors($response['headers']['x-open-runtimes-log-id']));
+    }
+
+    public function testResponseChunkedComplex(): void
+    {
+        $body = '';
+        $chunks = 0;
+        $response = Client::execute(body: 'Hello', headers: ['x-action' => 'responseChunkedComplex'], callback: function($chunk) use(&$body, &$chunks) {
+            $body .= $chunk;
+            $chunks += 1;
```
Meldiron, line 487: Not sure if this is enough. Would be more confident knowing it took ~1s between each chunk, to ensure sleeping in-between works, and you still get all data, and you get it gradually (not all chunks once at the end)

tests/Base.php
```diff
@@ -462,6 +462,85 @@ public function testDeprecatedMethods(): void
...
+    public function testResponseChunkedErrorStartDouble(): void
+    {
+        $body = '';
+        $chunks = 0;
+        $response = Client::execute(body: 'Hello', headers: ['x-action' => 'responseChunkedErrorStartDouble'], callback: function($chunk) use(&$body, &$chunks) {
+            $body .= $chunk;
+            $chunks += 1;
+        });
+
+        self::assertEquals(200, $response['code']);
```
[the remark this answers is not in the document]
Meldiron: Wrong, it's 200 becasue request started chunk response. In which case status code is set during start, when it's set to 200 because  at start all is good.

loks0n:

PR #278, opened 2024-07-09
Fix: Enforced headers

Meldiron:
Fix enforced headers in Go, and add test case in all runtimes.


Meldiron (commented):
Self-reviewed, looks good to me

vermakhushboo (approved):

PR #283, opened 2024-07-11
add node 22 runtime

liudonghua123:


byawitz:
Hi @liudonghua123, thanks for the PR.

We've added Node 22 using the latest runtimes structure, and you can check this PR #304.

PR #285, opened 2024-07-18
Feat: supervisord for auto-restarts

Meldiron:
- Fix tests (they were always passing)
- Add Supervisord to all runtiems
- Remove PM2 from Node
- Remove Denonn from Deno
- Update HTTP server to show better logs (for local development, as well as test case)
- Update tests for new action to exit process


tests/Base.php
```diff
@@ -623,6 +623,28 @@ function testEnforcedHeaders(): void
         self::assertEquals("248", $body['x-open-runtimes-custom']);
     }
 
+    /*
```
Meldiron, line 626: Uncomment this

ci/runtimes.toml
```diff
@@ -1,32 +1,32 @@
 [node]
 entry = "tests.js"
 versions = ["21.0", "20.0", "19.0", "18.0", "16.0", "14.5", "21.0-mjs", "20.0-mjs", "19.0-mjs", "18.0-mjs", "16.0-mjs", "14.5-mjs"]
-commands = { install = "npm install", start = "pm2 start src/server.js --no-daemon" }
+commands = { install = "npm install", start = "supervisord -c /etc/supervisord.conf -u root" }
```
byawitz, line 4: Maybe in the future, it means we can get rid of the `start` command altogether

byawitz (approved):

Meldiron (commented):
Self-reviewed file by file, no leftovers.

abnegate:
Why do we use supervisor instead of a restart policy for the container? Using supervisor inside containers is [discouraged by Docker](https://docs.docker.com/config/containers/start-containers-automatically/#using-a-process-manager-inside-containers)

Meldiron:
This PR is on hold. We will have auto-restart on executor level, and if that's not enough, then we can consider this PR again.

Meldiron:
Other relevent changes was merged here: https://github.com/open-runtimes/open-runtimes/pull/289

Closing this now

PR #288, opened 2024-07-25
Feat: Formatter

Meldiron:
Add formatter, and automated tests.


runtimes/node/versions/latest/src/server.js
```diff
@@ -2,233 +2,289 @@ const micro = require("micro");
...
+	);
 
-    try {
-        await action(logger, req, res);
-    } catch (e) {
-        logger.write(e, Logger.TYPE_ERROR);
+	try {
+		await action(logger, req, res);
+	} catch (e) {
+		logger.write(e, Logger.TYPE_ERROR);
```
christyjacob4, line 16: The default indentation for js seems to be a bit too much . Can we reduce it?

christyjacob4 (approved):
Looks good. I think we should check the default indentation for JS and reduce it. TS looks fine 👍

runtimes/node/versions/latest/src/server.js
Meldiron: It's same for Bun. It's 1 tab. I think it's okay, that's same as 4 spaces. And There isn't clear winner between 2 and 4 spaces.

Meldiron: I guess GitHub renders tab badly. Switched to spaces in biome.js config file

christyjacob4 (approved):

PR #290, opened 2024-07-29
Chore: Remove process manager

Meldiron:
Process manager wont be needed because executor will have restart policy: https://github.com/open-runtimes/executor/pull/90


Meldiron:
Fun fact, it also removes a deprecation warning 😁
CleanShot 2024-07-30 at 13 27 56@2x

christyjacob4 (approved):

PR #294, opened 2024-08-16
Feat: Support spread operator in logs

Meldiron:
Allows multiple params in log and write methods.

Runtimes:

- [x] Bun
- [x] CPP (Doesnt seem supported by language)
- [x] Dart (Doesnt seem supported by language)
- [x] Deno
- [x] Dotnet (Not supported by language)
- [x] Go
- [x] Java
- [x] Kotlin
- [x] Node
- [x] PHP
- [x] Python
- [x] Ruby
- [x] Swift (Not supported by language)

### DIFF
--- a/runtimes/go/versions/latest/src/go.mod
+++ b/runtimes/go/versions/latest/src/go.mod
@@ -4,5 +4,5 @@ go 1.22.5
 
 replace openruntimes/handler v0.0.0 => /usr/local/build
 
-// require openruntimes/handler v0.0.0
-require github.com/open-runtimes/types-for-go/v4 v4.0.6-0.20240818164233-b4d5d5df2b30
+require openruntimes/handler v0.0.0
+require github.com/open-runtimes/types-for-go/v4 v4.0.6
--- a/runtimes/go/versions/latest/src/go.mod
+++ b/runtimes/go/versions/latest/src/go.mod
@@ -4,5 +4,5 @@ go 1.22.5
 
 replace openruntimes/handler v0.0.0 => /usr/local/build
 
-require openruntimes/handler v0.0.0
-require github.com/open-runtimes/types-for-go/v4 v4.0.5
+// require openruntimes/handler v0.0.0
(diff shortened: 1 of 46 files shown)


runtimes/go/versions/latest/src/go.mod
```diff
@@ -4,5 +4,5 @@ go 1.22.5
 
 replace openruntimes/handler v0.0.0 => /usr/local/build
 
-require openruntimes/handler v0.0.0
-require github.com/open-runtimes/types-for-go/v4 v4.0.5
+// require openruntimes/handler v0.0.0
```
Meldiron, line 7: Remove?!? Seems unneeded?!? This is so good, it was always hussle with this when devleoping

runtimes/go/versions/latest/src/go.mod
```diff
@@ -4,5 +4,5 @@ go 1.22.5
...
+require github.com/open-runtimes/types-for-go/v4 v4.0.6-0.20240818164233-b4d5d5df2b30
```
Meldiron, line 8: Release version

tests/resources/functions/go/latest/go.mod
```diff
@@ -4,7 +4,7 @@ go 1.22.5
 
 require (
 	github.com/go-resty/resty/v2 v2.11.0
-	github.com/open-runtimes/types-for-go/v4 v4.0.5
+	github.com/open-runtimes/types-for-go/v4 v4.0.6-0.20240818164233-b4d5d5df2b30
```
Meldiron, line 7: Release version

christyjacob4 (approved):

abnegate:
@Meldiron This is supported in the some of the other languages:

C++: `void log(const std::string &message, ...)`
.NET: `public void Log(params object[] message)`
Swift: `func log(_ message: Any...)`

Might be because of a terminology mix-up: these are called variadic parameters, spread operator is what allows you to expand an array into it's individual items. Similar, but used in different contexts.

PR #304, opened 2024-09-27
Feat latest runtimes

byawitz:
Adding 
- Dart 3.5 
- Java 22 
- Kotlin 2.0 
- Node 22 
- Swift 5.10


runtimes/kotlin/versions/2.0/Dockerfile
```diff
@@ -0,0 +1,6 @@
+# syntax = devthefuture/dockerfile-x:1.4.2
+FROM eclipse-temurin:21-jdk-alpine
```
Meldiron, line 2: Dont this need to be JDK 22? Just guessing, from seeing latest Java is 22

PR #308, opened 2024-10-01
Fix: Version consistency

Meldiron:
- Deno used 1.41 instead of 1.40, fixed
- Lock Go, PHP, Python, Ruby, Node to specific patch versions


runtimes/deno/versions/1.40/Dockerfile
```diff
@@ -1,5 +1,5 @@
 # syntax = devthefuture/dockerfile-x:1.4.2
-FROM denoland/deno:alpine-1.41.3
+FROM denoland/deno:alpine-1.40.5
```
Meldiron, line 2: Intentional downgrade

christyjacob4 (approved):

PR #309, opened 2024-10-07
Feat: Improve python performance

Meldiron:
- Use gunicorn with proper thread setting per-code
- Remove unnessessary async

### DIFF
--- a/runtimes/python/versions/latest/src/server.py
+++ b/runtimes/python/versions/latest/src/server.py
@@ -10,6 +10,7 @@ import os
 import importlib
 import urllib.parse
 import asyncio
+import sys
 
 app = Flask(__name__)
 app.config["MAX_CONTENT_LENGTH"] = 20 * 1024 * 1024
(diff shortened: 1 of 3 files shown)


loks0n (approved):

runtimes/python/versions/latest/src/server.py
```diff
@@ -162,30 +145,25 @@ async def execute(context):
...
+async def handler(request) -> web.Response:
     logger = Logger(
         request.headers.get("x-open-runtimes-logging", ""),
         request.headers.get("x-open-runtimes-log-id", ""),
     )
 
     try:
         return await action(logger, request)
+    except web_exceptions.HTTPClientError as e:
+        return e
```
Meldiron, line 157: What is this doing exactly, do we have specific use case?

I am worried it could expose some data to client that's meant for developer.

Meldiron (commented):
LGTM

PR #312, opened 2024-10-13
Install server dependencies for python in dockerfile

vermakhushboo:
Install server dependencies for python in dockerfile instead of build step


Meldiron:
Closing as we already install server dependencies in Dockerfile in main branch

PR #313, opened 2024-10-13
fix: build server dependencies during docker build

christyjacob4:
- Install server dependencies for python during image build and copy them during start 

The cp does add 0.06s of latency to cold starts. I am trying to see if there is a better way
 
```bash
2024-10-13 15:39:54 real        0m 0.06s
2024-10-13 15:39:54 user        0m 0.00s
2024-10-13 15:39:54 sys 0m 0.04s
```

We have to use this approach since we cannot modify any of the build scripts as it would need a redeployment of the function.

Following this release, we can update the runtime such that this process happens in the build step.


christyjacob4:
This is a second approach which would work without introducing latency to the startup https://github.com/open-runtimes/open-runtimes/pull/315

vermakhushboo (approved):

runtimes/python/python.dockerfile
```diff
@@ -1,3 +1,9 @@
 ENV OPEN_RUNTIMES_ENTRYPOINT=main.py
+
 COPY requirements.txt .
-ENV FLASK_APP="/usr/local/server/src/server.py"
+
+# Create a virtual environment for the server dependencies
+RUN python3 -m venv server-env
+RUN server-env/bin/pip install --no-cache-dir -r requirements.txt
+
+ENV FLASK_APP="/usr/local/server/src/server.py"
```
eldadfux, line 9: Do we still need this?

eldadfux (approved):

runtimes/python/python.dockerfile
christyjacob4: We need the entry point - it is still server.py 
But the env variable name needs to be changed

christyjacob4: removed 👍

PR #315, opened 2024-10-13
Fix python build

christyjacob4:


christyjacob4:
We did a QA and discovered that this approach will not work. 

This is because the build tar.gz has already been created and stored in S3. As a result, our changes to the Dockerfile will not be included in it unless we have a rebuild

Meldiron:
Another possible approach, simply source 2 different folders:

CleanShot 2024-10-17 at 10 46 07@2x

Meldiron:
my approach wont work, you cannot source 2 virtual envs

Meldiron:
Closing this PR, since new ideas dont work, and packaging is already moved to dockerfile (server dependencies not included in build file)

PR #321, opened 2024-10-18
Feat: RC support

Meldiron:
- [x] QA ensuring my approach will work during release

CleanShot 2024-10-18 at 15 27 44@2x

CleanShot 2024-10-18 at 15 27 59@2x


loks0n:

PR #334, opened 2024-11-19
Chores: Improve tests structure

Meldiron:
- [x] Re-write tests to adapter approach for serverless, static, and websites


runtimes/dart/versions/latest/helpers/build-flutter.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
```
Meldiron, line 1: Safe to ignore this file, separate PR is adding proper flutter implementation. This was done as test for new tests structure

vermakhushboo (approved):

tests/Serverless/Php.php
christyjacob4, line 1: why aren't we using PHP?

tests/Serverless/PythonMl.php
christyjacob4, line 1: PythonML

christyjacob4 (approved):

PR #348, opened 2024-12-21
Feat: SSR runtime

Meldiron:
Separate SSR functionality into separate SSR runtime.

Allowing even more addition in future (expected already with handler-approach to SSR builds, and improved logging)

Also ensuring to keep Node runtime simple and unrelated to web frameworks, as time goes on


Meldiron (commented):

PR #349, opened 2024-12-21
Fix: React logs

Meldiron:
Log tests in Remix and Next.js were failing, they weren't implemented properly.

To fix we need to switch to custom HTTP server that uses SSR handler.

This PR does exactly that, resolving all tests to become green.


Meldiron (commented):

PR #350, opened 2024-12-22
Feat: Adapter-approach in frameworks

Meldiron:
Where possible, I applied adapter approach.

This makes cold-start faster, because we don't need proxy. It also gives more control over SSR server, which will be helpful in near future for solving concurrent logs


Meldiron (commented):

PR #353, opened 2024-12-25
Feat: Analog SSR  handler

Meldiron:
Uses handler approach to avoid the need to use proxy with Analog SSR framework.
Makes it faster, and easier to add custom logic (like logging concurrency in near future)


Meldiron (commented):

PR #354, opened 2024-12-26
Feat: Angular handler approach

Meldiron:
Angular was last remaining that used proxy. Now all frameworks use handler approach, giving us great possibilities with login reliability and speed.


Meldiron (commented):

PR #355, opened 2024-12-27
Chore: Remove SSR proxy

Meldiron:
SSR Proxy is no longer needed in any of the web frameworks


Meldiron (commented):

PR #366, opened 2025-02-24
Create Dockerfile

ItzNotABug:


runtimes/flutter/versions/3.27.3/Dockerfile
Meldiron, line 1: file should be just `3.27`

ci/runtimes.toml
```diff
@@ -39,7 +39,7 @@ tools = "dart --version"
 test = "Serverless/Dart.php"
 
 [flutter]
-versions = ["3.24"]
+versions = ["3.24", "3.27"]
```
Meldiron, line 42: Other go from latest to oldest, I think there is some logic that takes first from list and expect latest version. Left switch order


```suggestion
versions = ["3.27", "3.24"]
```

Meldiron (approved):

PR #371, opened 2025-03-12
Fix: Hidden files (attept 2)

Meldiron:
Removing `*` allows hidden  files (like .vitepress) to be copied during build.

It's okay to copy hidden files. They are not private, just meant to be hidden from user. Good example is `.prettierrc.json` - perfectly fine to copy, and even required in some CI/CD scenarios.


Meldiron:
Closing in favour of https://github.com/open-runtimes/open-runtimes/pull/372

PR #374, opened 2025-03-19
Chore throw error on empty source

vermakhushboo:
Ensure that the source directory exists and is not empty before proceeding with the build.


helpers/before-build.sh
```diff
@@ -4,6 +4,12 @@ set -e
 
 echo "Preparing for build ..."
 
+# Check if source directory exists and has files
+if [ ! -d "/mnt/code" ] || [ -z "$(ls -A /mnt/code 2>/dev/null)" ]; then
+    echo "ERROR: No source code found. Please ensure your source directory exists and isn't empty."
```
Meldiron, line 9: Let's share the scenario with Chen, and ask what's the best copy to show to developer in logs

PR #378, opened 2025-03-27
feat: workspace runtime

ChiragAgg5k:
Adds a new runtime called the `workspace`, which utilizes the synapse library (by appwrite) to have an environment for user's workspace on the browser.


eldadfux (changes requested):

.github/workflows/publish.yaml
```diff
@@ -63,6 +63,13 @@ jobs:
             VERSION: "14.5"
             IMAGE: openruntimes/node:v4-14.5
             ARCH: "linux/amd64"
+        
+          # Workspace
+          - ID: workspace-22.js
```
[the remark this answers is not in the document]
Meldiron: Agree, good example of this is `static` runtime, which follows our versioning, and simply says version 1. We could version it same as underlying Rust http server, but there is no benefit in that.

runtimes/workspace/versions/latest/src/server.js
```diff
@@ -0,0 +1,226 @@
...
+          case "push":
+            result = await git.push();
+            break;
+          default:
+            throw new Error("Invalid git operation");
+        }
+        return result;
+      },
+
+      codeStyle: async (message) => {
```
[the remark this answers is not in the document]
ChiragAgg5k: done

runtimes/workspace/versions/latest/src/proxy.js
```diff
@@ -0,0 +1,87 @@
+const { createProxyServer } = require("http-proxy");
```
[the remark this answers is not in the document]
ChiragAgg5k: yes it was present in all the other runtimes, maybe matej can answer this. i do see its called in one of the scripts

runtimes/workspace/versions/latest/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "type": "commonjs",
+  "main": "server.js",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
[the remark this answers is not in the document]
ChiragAgg5k: its used in node runtime as well

runtimes/workspace/versions/latest/src/server.js
```diff
@@ -0,0 +1,226 @@
...
+    };
+
+    Object.keys(router).forEach((type) => {
+      synapse.onMessageType(type, async (message) => {
+        console.log("Message received:", message);
+        try {
+          const result = await router[type](message);
+          if (result !== null) {
+            synapse.send(`${type}Response`, {
+              requestId: message.requestId,
```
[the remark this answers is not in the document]
ChiragAgg5k: its generated by the ws library automatically

.github/workflows/publish.yaml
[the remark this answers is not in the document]
ChiragAgg5k: sure 👍🏻

tests/Workspace.php
```diff
@@ -0,0 +1,419 @@
+<?php
+
+namespace Tests;
+
+use PHPUnit\Framework\TestCase;
+use Utopia\WebSocket\Client as WebsocketClient;
+
+use function Swoole\Coroutine\run;
+
+class Workspace extends TestCase
```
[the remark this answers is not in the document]
ChiragAgg5k: still WIP to be added to synapse

runtimes/workspace/versions/latest/src/server.js
```diff
@@ -0,0 +1,389 @@
...
+      }
+    }
+
+    // Health endpoint
+    if (url === "/health") {
+      return send(res, 200, { success: true, data: "OK" });
+    }
+
+    // Terminal endpoints
+    if (url.startsWith("/terminal/")) {
```
eldadfux, line 295: Not sure we want to have the server as specific or REST compatible like we usually do, it will require a bit more work for us now. Maybe we can just have a single POST route for all operations, maybe even as simple as `POST /` without having a namespace for each route like `/terminal`, `/fs`, etc
Where we pass `service` and `operation` as params.

eldadfux: The current approach is just not REST and not-REST somewhere in the middle which is confusing and still requires maintenance.

tests/Client.php
```diff
@@ -30,7 +31,7 @@ public static function execute($body = '', $url = '/', $method = 'POST', $header
 
         $responseHeaders = [];
         $optArray = [
-            CURLOPT_URL => 'http://localhost:' . self::$port . $url,
+            CURLOPT_URL => 'http://' . self::$host . ':' . self::$port . $url,
```
eldadfux, line 34: why is this static instead of instance based? We could probably use utopia/fetch, if this is too big of a change we can just fix the static stuff and have a PR later for switching to the fetch lib.

tests/Workspace/HTTPTest.php
```diff
@@ -0,0 +1,327 @@
...
+        $response = Client::execute(url: '/health');
+        $json = json_decode($response['body'], true);
+
+        $this->assertEquals(200, $response['code']);
+        $this->assertTrue($json['success']);
+        $this->assertEquals('OK', $json['data']);
+    }
+
+    public function testTerminalOperations(): void
+    {
```
eldadfux, line 57: I think we can have better organization here. Maybe have a class per service, and a test per operation.

tests/Workspace/WebsocketsTest.php
```diff
@@ -0,0 +1,399 @@
+<?php
+
+namespace Tests\Workspace;
+
```
eldadfux, line 4: We could try to have a common class for all shared functionality between http and websocket tests, and then just have an extension class that just changes the protocol we use to fetch the data.

eldadfux (changes requested):

tests/Workspace/WebsocketsTest.php
ChiragAgg5k: all the functionality is technically common, but its hard to have a common class because of the amount of differences there are. It's not just the protocol used, but also the way we fetch the response, the way we run websocket code in a coroutine etc.

tests/Workspace/HTTPTest.php
ChiragAgg5k: hmm sure we can do this

tests/Client.php
ChiragAgg5k: we also define the ports statically. there is no constructor defined for the Client class yet.

ChiragAgg5k: if we plan to use utopia/fetch, lets do that in a future PR and implement it across the repo

runtimes/workspace/versions/latest/src/server.js
ChiragAgg5k: i agree we the current approach is not very REST and we probably should have something REST specific... 

but i dont know if i like the idea of one endpoint `POST /`

tests/Workspace/WebsocketsTest.php
ChiragAgg5k: did try some things but its taking too long and leading to more and more errors. maybe we can have this in future. 

Screenshot 2025-04-07 at 7 50 52 PM

tests/Workspace/HTTPTest.php
```diff
@@ -0,0 +1,97 @@
+<?php
+
+namespace Tests\Workspace;
+
+use Tests\Client;
+
+class HTTPTest extends Workspace
```
Meldiron, line 7: HTTPTest sounds too generic, could be used for any of the runtimes.

Notice patterm we use - Base, then Serverless, then Node.
Let's follow same here - Let's have Websocket  tests, where we simply test some simple connect&ping&pong&disconnect. Then lets have Workspace that extends Websocket, where we test all methods we defined in server.js of workspace.

tests/Workspace/WebsocketsTest.php
```diff
@@ -0,0 +1,136 @@
...
+        run(function () {
+            $this->client->connect();
+            parent::testTerminalOperations();
+            $this->client->close();
+        });
+    }
+
+    public function testFilesystemOperations(): void
+    {
+        run(function () {
```
Meldiron, line 91: Why do we need all those runs? Is that something swoole specific? Can we avoid it, or do it only in 1 place?

tests/Workspace/Workspace.php
```diff
@@ -0,0 +1,322 @@
...
+            'type' => 'fs',
+            'operation' => 'deleteFolder',
+            'params' => [
+                'folderpath' => 'testfolder'
+            ]
+        ]);
+        $this->assertTrue($response['success']);
+
+        /**
+         * Test for FAILURE
```
Meldiron, line 138: Let's have more failure tests.
cannot delete non-existing folder & file, cannot create duplicate file/folder (cannot create file of same name as folder too).

Lets validate all other tests if they can have more failure tests

ci_tests.sh
```diff
@@ -81,4 +90,19 @@ docker run -d --name open-runtimes-test-serve-teritary -v $(pwd)/resources:/mnt/
...
+    --network executor_runtimes \
+    -w /app \
+    -e RUNTIME_NAME="$RUNTIME" \
+    -e RUNTIME_VERSION="$VERSION" \
+    -e OPEN_RUNTIMES_SECRET="test-secret-key" \
+    -e OPEN_RUNTIMES_ENTRYPOINT="$ENTRYPOINT" \
+    phpswoole/swoole:5.1.2-php8.3-alpine \
+    sh -c "apk update && apk add docker-cli zip unzip && composer install --profile --ignore-platform-reqs && vendor/bin/phpunit --configuration phpunit.xml tests/$TEST_CLASS" --process-isolation --debug
+else
+  RUNTIME_NAME="$RUNTIME" RUNTIME_VERSION="$VERSION" OPEN_RUNTIMES_SECRET="test-secret-key" OPEN_RUNTIMES_ENTRYPOINT="$ENTRYPOINT" vendor/bin/phpunit --configuration phpunit.xml tests/$TEST_CLASS --debug
```
Meldiron, line 107: Let's avoid huge if/else.
Instead, let's do everything inside docker. Only thing we will need to change is image.
i would also not define it here, I would define it in .toml file as `testImage`. You can use standard php image for all, and swoole for workspace specifically

ci_tests.sh
```diff
@@ -19,22 +22,28 @@ if [ "$VERSION" = "$LATEST_VERSION" ]; then
...
 
-    cd "tests/resources/functions/$RUNTIME_FOLDER"
-    docker run --rm --name open-runtimes-formatter -v $(pwd):/mnt/code:rw open-runtimes/test-runtime sh -c "cd /mnt/code && $FORMATTER_PREPARE && $FORMATTER_CHECK"
-    cd ../../../../
+    if [ -d "./tests/resources/functions/$RUNTIME_FOLDER" ]; then
+        cd "tests/resources/functions/$RUNTIME_FOLDER"
+        docker run --rm --name open-runtimes-formatter -v $(pwd):/mnt/code:rw open-runtimes/test-runtime sh -c "cd /mnt/code && $FORMATTER_PREPARE && $FORMATTER_CHECK"
+        cd ../../../../
+    else
+        echo "Skipping formatter for tests as runtime folder $RUNTIME_FOLDER does not exist"
```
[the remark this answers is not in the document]
ChiragAgg5k: it was added because there are no files in /functions/ folder for workspace runtime

tests/Workspace/WebsocketsTest.php
ChiragAgg5k: technically we should not require runs but there is some bug in utopia-php/websockets that does not allow the function to run without swoole.

i can look into making it in one place but the fix should be made there eventually

tests/Workspace/HTTPTest.php
ChiragAgg5k: But we still need HTTP based tests... how would we fit that?

tests/Workspace/Workspace.php
ChiragAgg5k: sure will add more

ci_tests.sh
ChiragAgg5k: sooo change all the tests to run in docker and not locally 👀
ok 👍🏻

tests/Workspace/WebsocketsTest.php
ChiragAgg5k: done this now.

tests/Workspace/HTTPTest.php
ChiragAgg5k: i have updated the file naming for now 👍

tests/Client.php
```diff
@@ -50,7 +51,7 @@ public static function execute($body = '', $url = '/', $method = 'POST', $header
         ];
 
         if($body !== NULL) {
-            $optArray[CURLOPT_POSTFIELDS] = \is_array($body) ? \json_encode($body, JSON_FORCE_OBJECT) : $body;
+            $optArray[CURLOPT_POSTFIELDS] = \is_array($body) ? \json_encode($body) : $body;
```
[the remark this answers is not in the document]
Meldiron: If it doesnt break existing tests, fine by me.

runtimes/workspace/versions/latest/src/server.js
```diff
@@ -0,0 +1,320 @@
...
+    const { operation, params } = message;
+    let result;
+
+    switch (operation) {
+      case "init":
+        result = await git.init();
+        break;
+      case "addRemote":
+        result = await git.addRemote(params.name, params.url);
+        break;
```
TorstenDittmann, line 122:
```suggestion
      case "init":
        return git.init();
      case "addRemote":
        return git.addRemote(params.name, params.url);
```

no need for the result param 👍🏻

PR #390, opened 2025-05-30
feat(bun): add bun 1.2

verydanny:
### What
Add bun 1.2 to the runtimes.toml and the Dockerfile.

### Why
There's a particularly annoying issue with JSX that prevents the use of things like { h }, jsx-lite, and others. It basically makes bun ignore tsconfig.json. It was fixed here: https://github.com/oven-sh/bun/issues/3768

We have to upgrade to Bun 1.2+ to fix this

### Issues
https://github.com/oven-sh/bun/issues/3768

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -347,6 +347,11 @@ jobs:
             ARCH: "linux/amd64,linux/arm64"
 
           # Bun
+          - ID: bun-1.2
+            RUNTIME: bun
+            VERSION: "1.2"
+            IMAGE: openruntimes/bun:v5-1.2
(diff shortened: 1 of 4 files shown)


PerryFinn:
The check has not passed.

hlysine:
The check failures are from dart, which cannot be related to the bun runtime. Can we get this merged soon? Bun 1.2 has been out for over half a year and is really needed given how buggy Bun 1.1 was.

itea-dev:
Last bun version is now 1.3.1

loks0n:
Thanks! Bun is now maintained under the rolling runtimes/bun/latest layout, which is past 1.2 — superseding this PR. Closing.

PR #398, opened 2025-07-02
Add shfmt check to CI/CD

vermakhushboo:


loks0n:
Thanks for the contribution! A shfmt check now runs in CI (.github/workflows/test.yaml), so this is covered. Closing as implemented.

PR #400, opened 2025-07-07
feat: telemetry

loks0n:


ci/runtimes.toml
```diff
@@ -123,7 +149,7 @@ test = "Serverless/Bun.php"
 versions = ["1"]
 commands = { install = "bash build.sh", start = "bash helpers/server.sh" }
 formatter = { prepare = "echo No formatter preparations.", check = "echo No formatter checker.", write = "echo No formatter writer." }
-tools = "static-web-server --version"
+tools = "caddy --version"
```
Meldiron, line 152: I think we need to stick to static-web-server here

runtimes/node/versions/latest/src/server-next-js.mjs
```diff
@@ -1,11 +1,22 @@
...
 import express from "express";
+import { readFileSync } from "fs";
 import { onInit, getPort, getHost, onAction, onError } from "./ssr/helpers.mjs";
 
 console.log("Next.js server starting ...");
 
 const app = express();
+
+app.use((req, res, next) => {
+  if (req.headers["x-open-runtimes-timings"]) {
```
Meldiron, line 12: If we use path instead, we should not need caddy in static runtime

helpers/before-start.sh
```diff
@@ -3,18 +3,29 @@
 set -e
 shopt -s dotglob
 
+# Prepare telemetry
+mkdir -p /usr/local/telemetry/
+[ -d "/mnt/telemetry" ] && cp /mnt/telemetry/* /usr/local/telemetry/ 2>/dev/null || true
```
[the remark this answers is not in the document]
loks0n: Just copied the setup we have for the existing mount.

We can instead mount as readwrite and just serve that directly

helpers/before-start.sh
```diff
@@ -3,18 +3,26 @@
...
 
 # Extract code from mounted volume to function folder
+start_ms=$(date +%s%3N)
 if [ -f /mnt/code/code.tar ]; then
     tar -xf /mnt/code/code.tar -C /usr/local/server/src/function
 elif [ -f /mnt/code/code.tar.gz ] || [ -f /mnt/code/code.gz ]; then
     tar -zxf /mnt/code/code.tar.gz -C /usr/local/server/src/function
 else
-    echo "[90m$(date +[%H:%M:%S]) [31m[[00mopen-runtimes[31m][97m Code archive not found. [0m"
+    echo -e "[90m$(date +[%H:%M:%S]) [31m[[00mopen-runtimes[31m][97m Code archive not found. [0m"
```
Meldiron, line 18: What does the -e do here? I recall different version here, when you replaces ESC characters with \e or something similar.

If thats the case, lets remove this -e, and let me know, and Ill setup a separate task for doing this to all relevant echo commands

tests/CSR/Flutter.php
```diff
@@ -26,4 +26,9 @@ public function testEmptyServerSecret(): void
     {
         $this->assertTrue(true);
     }
+
+    public function testTimings(): void
```
Meldiron, line 30:
```suggestion
    // Timings currently run during cold-start, and Flutter's runtime static cold-start is not used (static recommended instead)
    public function testTimings(): void
```

tests/resources/telemetry/timings.txt
Meldiron, line 1: let's add this directory to gitignore

runtimes/flutter/versions/latest/helpers/before-start.sh
```diff
@@ -5,7 +5,12 @@ shopt -s dotglob
 
 echo "Preparing for start ..."
 
+# Prepare telemetry
+mkdir -p /usr/local/telemetry/
+[ -d "/mnt/telemetry" ] && cp /mnt/telemetry/* /usr/local/telemetry/ 2>/dev/null || true
```
Meldiron, line 10:
```suggestion
```

runtimes/node/versions/latest/src/server-analog.mjs
```diff
@@ -1,10 +1,21 @@
 import { handler } from "./server/index.mjs";
 import express from "express";
+import { readFileSync } from "fs";
```
Meldiron, line 3: In places we used synchronous methods in Deno/Node/Bun, lets please use async method instead?

runtimes/node/versions/latest/src/server-next-js.mjs
```diff
@@ -1,11 +1,22 @@
...
 const app = express();
+
+app.use((req, res, next) => {
+  if (req.path === "/__opr/timings") {
+    const timings = readFileSync("/mnt/telemetry/timings.txt", "utf8");
+    res.setHeader("content-type", "text/plain; charset=utf-8");
+    return res.status(200).send(timings);
+  }
+  next();
+});
```
Meldiron, line 18: Notice just below this, we have app.use(onInit), onInit coming from helpers.mjs.
Let's make this into helper function too, and do app.use(telemetryEndpoint) or such.

This makes it easier in future when we need to update the code, it would be in just 1 place for all ssr frameworks

runtimes/python/versions/latest/src/server.py
```diff
@@ -135,7 +136,7 @@ async def execute(context):
     )
     if (
         not resp.headers["content-type"].startswith("multipart/")
-        and not "charset=" in resp.headers["content-type"]
+        and "charset=" not in resp.headers["content-type"]
```
Meldiron, line 139: As per AI, those both resolve in same logic, so all good

tests/resources/telemetry/timings.txt
loks0n: It needs to exist so it can be mounted

runtimes/node/versions/latest/src/server-analog.mjs
loks0n: it's only this one using sync, swapped for fs/promise

runtimes/flutter/versions/latest/helpers/before-start.sh
loks0n: removed

ci/runtimes.toml
```diff
@@ -11,7 +26,7 @@ entry = "tests.ts"
 versions = ["2.0", "1.46", "1.40", "1.35", "1.24", "1.21"]
 commands = { install = "deno cache tests.ts", start = "bash helpers/server.sh" }
 formatter = { prepare = "echo \"Using native formatter\"", check = "deno fmt --check", write = "deno fmt" }
-tools = "deno --version" # TODO: NPM, PNPM and Yarn dont work in Deno alpine currently: https://github.com/denoland/deno_docker/issues/373
+tools = "deno --version"                                                                                    # TODO: NPM, PNPM and Yarn dont work in Deno alpine currently: https://github.com/denoland/deno_docker/issues/373
```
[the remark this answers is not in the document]
loks0n: this is from a toml auto format

helpers/before-start.sh
loks0n: not neccessary, removing

.github/workflows/test.yaml
```diff
@@ -2,6 +2,10 @@ name: "Testing OpenRuntimes"
 
 on: [ pull_request ]
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
```
loks0n, line 8: Without this merge tests are very annoying, we run out of runners and you have to cancel manually

Meldiron: ah, nice! I had task for it for next cycle, amazing it's done already

PR #404, opened 2025-07-11
fix: timings precision

loks0n:


Meldiron (approved):
LGTM, let's update tests if possible

PR #413, opened 2025-07-28
Feat: rework node image versionning

### LINKED ISSUE #412: Add support for Angular 20
As shown below Angular 20 needs either node 20.19 but you have 20.17, or node 22.12 but you have 22.9.

Image

Could you please update the images so we can use open runtimes for angular 20.
(trying to build and deploy it on Appwrite Sites)

ibaraki-douji:
This is a rework of the node images names as before only v14.5 was following the node version. all others vXX.0 runtime where not node minor XX.0 but something else .

This PR is here to make all node images work the same way with versions :
- `node-<major>` : use the `node:<major>-alpine` image (may change to keep it as the latest major)
- `node-<major>.0` : use the `node:<major>.0.0-alpine<alpVersion>` (always include a first minor at 0)
- `node-<major>.<minor>`: use the `node:<major>.<minor>.<patch>-alpine<alpVersion>` (can be created when new node images are created, should never change as thoses use the whole version with patches)

It should also fix #412 as i added the latest node 20/22/24.


loks0n:
Thanks for the proposal! The repo has since moved to a rolling versions/latest layout for Node (and all runtimes), which supersedes this approach. Closing.

PR #424, opened 2025-09-12
Add Flutter 3.35.3 runtime

DH-555:


ci/runtimes.toml
```diff
@@ -65,7 +65,7 @@ tools = "dart --version"
 test = "Serverless/Dart.php"
 
 [flutter]
-versions = ["3.32", "3.29", "3.27", "3.24"]
+versions = ["3.35.3", "3.32", "3.29", "3.27", "3.24"]
```
loks0n, line 68: can we just call it 3.35? we generally just bump patch versions

loks0n (changes requested):

ci/runtimes.toml
DH-555: Solved

PR #438, opened 2025-10-26
cicd: Add Dart 3.9 support in publish workflow

romaingyh:
Adds Dart 3.9 in publish workflow to publish the image on Docker Hub


loks0n:
Dart 3.9 is now in the tree and published, so this is covered. Thanks for the PR — closing as implemented.

PR #444, opened 2025-11-05
Feat: Fresh support

Meldiron:


Meldiron:
We should instead add generic support to any framework

PR #447, opened 2025-11-24
Add DB runtimes

abnegate:


abnegate:
Closing this old raw database runtime proposal because it sits outside the shipping Cloud and Edge dedicated-database architecture. If runtime-facing database support is pursued later, it needs a fresh design against the current architecture; this head must not be merged.

PR #448, opened 2025-11-25
Feat: Bun array header support

Meldiron:


runtimes/bun/versions/latest/src/server.ts
```diff
@@ -215,27 +215,41 @@ const action = async (logger: Logger, request: any) => {
...
+          responseHeaders.append(header.toLowerCase(), value);
+        }
+      } else {
+        responseHeaders.append(header.toLowerCase(), output.headers[header]);
+      }
+    }
+  }
 
-    responseHeaders[header.toLowerCase()] = output.headers[header];
+  for (const [key, _value] of responseHeaders) {
```
hmacr: We're filtering all `x-open-runtimes-` header here but adding `x-open-runtimes-log-id` later, is that expected?

PR #462, opened 2026-02-17
[Dart] Add image type check for content type encoding

Namaneo:
Add an extra check to not add `charset=utf-8` when the Content-Type is an image.

I ran into an image-related issue on a web app, and noticed in the Chrome console that this charset was incorrectly added. This doesn't really make sense and breaks some processes, e.g. `CachedNetworkImage` in Flutter doesn't like it. Also I suspect this may break OpenGraph image queries.


Namaneo:
Applied Greptile suggestion.

PR #470, opened 2026-03-18
feat: Support for native req & res in Node

hmacr:
Closes SER-1157


runtimes/node/versions/latest/src/server.js
```diff
@@ -136,6 +136,25 @@ const action = async (logger, req, res) => {
       port,
       url,
       path,
+      asNative: function () {
```
Meldiron: lets discuss name

tests/Serverless/Javascript.php
```diff
@@ -0,0 +1,109 @@
...
+    {
+        if ($this->runtimeName === 'node' && str_starts_with($this->runtimeVersion, '16.0')) {
+            self::markTestSkipped('Node 16 does not support raw response object.');
+        }
+
+        $response = Client::executeStreaming(headers: ['x-action' => 'streamingResponse']);
+
+        self::assertEquals(200, $response['code']);
+        self::assertStringContainsString('chunk1', $response['body']);
+        self::assertStringContainsString('chunk2', $response['body']);
```
Meldiron: Lets reuse same logi we have in executor for build stremaing logs

PR #485, opened 2026-04-19
feat: add SSR sites support to bun + deno runtimes, share wrappers via runtimes/javascript/

ChiragAgg5k:
## Summary

Brings Appwrite Sites parity to two more JS runtimes and deduplicates the SSR plumbing along the way.

Before this PR, `runtimes/node` was the only runtime that could serve Astro / Angular / Analog / Next.js / Nuxt / Remix / SvelteKit / TanStack Start. `runtimes/bun` shipped only a functions server; `runtimes/deno` had no SSR wiring at all. All of the framework wrappers, bundle/env helpers, `ssr/system.mjs`, and `ssr/logger.mjs` lived under `runtimes/node/` and were only usable there.

This PR does three things at once:

1. **Adds full SSR support to bun** — working end-to-end for every supported framework.
2. **Adds SSR support to deno** — same contract, same shared tree, same CI coverage.
3. **Consolidates the JS runtimes** behind a shared tree at `runtimes/javascript/` that gets overlaid into any JS runtime's image at build time. Node, bun, and deno now share 25+ files verbatim.
4. **Wires CI to actually run every SSR framework test on all three runtimes** — not just node.

Depends on [#486](https://github.com/open-runtimes/open-runtimes/pull/486) (shared tree, merged) and [#487](https://github.com/open-runtimes/open-runtimes/pull/487) (logger + http-injection consolidation). This PR is based on #487 and brings bun + deno along.

---

## What's shared across JS runtimes — `runtimes/javascript/`

Every JS runtime (node, bun, deno) opts in via `shared = "javascript"` in its `ci/runtimes.toml` entry. `ci-runtime-prepare.sh` overlays `runtimes/javascript/*` onto the build context before the runtime-specific files, so everything lands at the same absolute paths in `/usr/local/server/` that callers already expect.

| File | What it does |
| --- | --- |
| `src/server-{analog,angular,astro,next-js,nuxt,remix,sveltekit,tanstack-start}.mjs` | 8 express-hosted framework wrappers. Copied next to user code at start time; their framework imports resolve from the user's `node_modules`. |
| `src/ssr/system.mjs` | `/__opr/health`, `/__opr/timings`, secret auth, safe-timeout header validation. |
| `src/ssr/logger.mjs` | Per-request logger with `console.*` override. Uses `AsyncLocalStorage` — stable in Node 16+ and native in bun and deno. |
| `src/ssr/http-injection.mjs` | `overrideEmit` / `wrapServer` / `wrapCreateServer` / `wrapHttp`. Applied by each runtime's `injections.*` entrypoint. |
| `src/ssr/deno.json` | Deno-only: import map for the shared preload + `"unstable": ["bare-node-builtins", "sloppy-imports"]`. Ignored by node and bun. |
| `helpers/<framework>/{bundle.sh,env.sh,server.sh}` | 8 × 3 build/serve helpers. `server.sh` sources `javascript-runner/env.sh` for runner + preload flag. |

---

## What stays per-runtime

| File | node | bun | deno |
| --- | --- | --- | --- |
| `helpers/javascript-runner/env.sh` | `OPR_JAVASCRIPT_RUNNER=node` + `NODE_OPTIONS='--import=…injections.mjs'` | `OPR_JAVASCRIPT_RUNNER=bun` + `BUN_OPTIONS=--preload=…injections.ts` | `OPR_JAVASCRIPT_RUNNER="deno run --allow-all --node-modules-dir=manual --config=…/deno.json --import=…/injections.ts"` |
| `src/ssr/injections.*` | `.mjs`, ~25 LOC — `import-in-the-middle` (ESM) + `require-in-the-middle` (CJS) + `node:module` `register()`. Node needs this because `--import` runs too late for CJS `require()` calls. | `.ts`, ~40 LOC — direct `http.Server.prototype.emit` patch at `--preload` time (beats user code). | `.ts`, ~35 LOC — same direct patch as bun; reads `Deno.env.get(...)` instead of `process.env[...]`. |
| `src/server.{js,ts}` | `micro` + Buffer | `Bun.serve` + Request/Response | existing oak-based deno server |
| `<runtime>.dockerfile` | + `modclean` for pruning | + `python3 make g++ gcc git` for native module builds | same tools **plus** `nodejs-current npm`, and `rm -f /usr/local/lib/libgcc_s.so*` to resolve the deno-glibc vs alpine-musl conflict |
| `src/nft.mjs`, `helpers/prepare-packing.sh` | NFT + modclean pruning | no-op | no-op |

---

### New
- `src/ssr/injections.ts` — bun `--preload` entrypoint. Imports `overrideEmit` / `wrapServer` / `wrapCreateServer` from shared `http-injection.mjs`. Patches `http.Server.prototype.emit` directly and wraps `http.createServer` / `http.Server` class.
- `helpers/javascript-runner/env.sh` — exports `OPR_JAVASCRIPT_RUNNER=bun` and `BUN_OPTIONS="--preload /usr/local/server/src/ssr/injections.ts"`.

### Modified
- `bun.dockerfile` — adds `python3`, `make`, `g++`, `gcc`, `git`. Deliberately avoids `gcompat` because bun's alpine image already ships glibc and the two conflict on `ld-linux-x86-64.so.2`.
- `package.json` — adds `express` and `srvx` so the shared wrappers' bare imports have a fallback at `/usr/local/server/node_modules/` when the user's own `node_modules` doesn't declare them (most Astro/SvelteKit/Nuxt projects don't). Mirrors node.
- `ci/runtimes.toml` `[bun]` — adds `shared = "javascript"`.

---

### New
- `src/ssr/injections.ts` — deno `--import` entrypoint. Byte-near-identical to bun's; only difference is `Deno.env.get(...)` instead of `process.env[...]`.
- `helpers/javascript-runner/env.sh` — exports `OPR_JAVASCRIPT_RUNNER="deno run --allow-all --node-modules-dir=manual --config=/usr/local/server/src/ssr/deno.json --import=/usr/local/server/src/ssr/injections.ts"`.

### Modified
- `deno.dockerfile`
  - Installs `nodejs-current`, `npm`, `pnpm`, `yarn`, `python3`, `make`, `g++`, `gcc`, `git` so the shared `npm install && npm run build && bash …/<framework>/bundle.sh` install flow works.
  - `rm -f /usr/local/lib/libgcc_s.so*` removes deno's bundled glibc-shim so node (musl-linked) can load the system libgcc from `/usr/lib`. Deno itself stays functional — its binary is statically linked.
  - On **deno 2.x only** (feature-probed via `deno install --help | grep -q -- '--allow-scripts'`), runs `deno install --node-modules-dir=auto --allow-scripts` against the shared `deno.json` so the SSR preload's `import "superjson"` resolves at runtime without network. Deno 1.x skips this step — its `deno install` is the CLI-script installer and SSR isn't viable there anyway (no `--import` flag).
- `ci/runtimes.toml` `[deno]` — adds `shared = "javascript"`, updates `tools` to reflect that npm/pnpm/yarn are now installed, removes the old TODO.

### Shared (lives in `runtimes/javascript/src/ssr/deno.json`)
- Import map for the shared preload's `superjson` plus the wrapper-level bare imports `express`, `srvx/node`, `@remix-run/express`, `next`. Under `--node-modules-dir=manual` deno doesn't walk up the module tree the way node/bun do, so these need to be pinned.
- `"unstable": ["bare-node-builtins", "sloppy-imports"]` — Nitro's compiled SSR bundles (tanstack nitro v2/v3, nuxt) `import "stream"` / `"url"` / `"http"` without the `node:` prefix. The flag teaches deno to treat them as `node:…`.

---

### 22 new `ci/runtimes.toml` entries (11 frameworks × bun + deno)

Naming convention: `<framework>_<runtime>` (astro_bun, astro_deno, …). Matches the existing `next-js-turbopack` variant convention and keeps the node-only check names (`astro`, `sveltekit`, …) stable for branch-protection purposes.

Each variant entry reuses the existing `tests/resources/functions/<framework>/` fixture via a new **`fixture = "<name>"`** field — avoids duplicating 22 fixture directories.

### Infrastructure plumbing

- `ci/index.ts` — adds the `fixture` field to the `Runtime` interface and passes it through the matrix as `FIXTURE`.
- `.github/workflows/test.yaml` — threads `FIXTURE` env into `ci_tests.sh`.
- `tests.sh` — reads `fixture` from TOML for local runs, plus a bug fix so `VERSION=null` collapses to `""` (matching the CI matrix's `cleanVersion('')`). Previously the formatter step diverged between local and CI.
- `ci_tests.sh`
  - Fixture copy now prefers `FIXTURE` when set, falls back to `RUNTIME`.
  - Skips the two extra SSR comparison builds (modclean-disabled baseline, NFT-enabled) for non-node runtimes — bun and deno ship the repo-root no-op `prepare-packing.sh` so there's nothing to compare.
  - Passes `ENFORCED_RUNTIME` through to the phpunit container.

### Test skips for node-only features

- `tests/SSR.php::testModclean`, `::testNft` — `markTestSkipped` when `ENFORCED_RUNTIME !== 'node'`. `modclean` and `@vercel/nft` pruning are wired only in node's `helpers/prepare-packing.sh`.
- `tests/SSR/NextJS.php::testNft` override — same guard (the NextJS version asserts NFT-is-skipped-for-nextjs, still node-only).

---

## Local e2e verification

Ran four representative variants end-to-end with `bash tests.sh …`:

| Variant | Result |
| --- | --- |
| `next-js_bun` | 19/19 (2 skipped: modclean+NFT) |
| `tanstack_start_nitro_v3_deno` | 19/19 (2 skipped) |
| `nuxt_deno` | 19/19 (2 skipped) |
| `tanstack_start_native_deno` | 19/19 (2 skipped) |

Plus a manual side-by-side of the Astro fixture served by both runtimes simultaneously (`http://localhost:3100` = bun, `http://localhost:3101` = deno). Identical output, identical OPR contract (`x-open-runtimes-log-id` header, `/__opr/health`, `/__opr/timings`, auth enforcement with `OPEN_RUNTIMES_SECRET`, safe-timeout behaviour).

## CI status

| Category | Node | Bun | Deno |
| --- | --- | --- | --- |
| Serverless (functions) | 17 versions | 4 versions | 8 versions |
| SSR (11 frameworks) | green | green | green (modulo the occasional `esm.sh` 5xx flake on deno 1.x — pre-existing, unrelated; those versions predate `--import` so SSR isn't wired there anyway) |

## Known limitations

1. **No dep pruning on bun/deno.** `helpers/prepare-packing.sh` falls back to the repo-root no-op, so `OPEN_RUNTIMES_CLEANUP=none`. Bun/deno SSR output tarballs are larger than node's until NFT/modclean are ported. Tests are gated accordingly.
2. **`OPEN_RUNTIMES_START_COMMAND` must match the runtime.** `BUN_OPTIONS` is only honoured by `bun`; `NODE_OPTIONS`/`--import` only by `node`; deno's flag set only by `deno`. A user-supplied start command that invokes a different binary bypasses the injection layer. Documented implicitly by living under each runtime's folder.
3. **Deno 1.x has no SSR.** The `--import` preload flag is 2.x-only, and `deno install --node-modules-dir` / `--allow-scripts` only exist in 2.x. The dockerfile step that pre-populates `/usr/local/server/node_modules/` skips itself on 1.x via a `--allow-scripts` help probe. SSR variants target deno 2.6.

## Commits

Implementation split across 18 commits for reviewability; commit messages explain each decision in detail. Key milestones:

1. `4343e120` — initial bun SSR support (before the shared tree existed)
2. `48870e56` / `89cab429` / `b0e8badf` / `ed3a7ed7` — bun migrates to the shared tree as #486 evolves
3. `07a53b93` — bun consumes shared logger + http-injection from #487
4. `8026ef18` — deno SSR scaffolding
5. `f79d640a` / `7468d625` / `f9c609e7` — deno preload fixes discovered during local e2e (deno.json, --allow-scripts probe, wrapper dep import map)
6. `f5983033` — ci wiring for SSR tests on bun + deno
7. `ee6966f3` / `07a5de50` — modclean / NFT skip guards for non-node
8. `64d46fb6` — bare-node-builtins unstable flag for deno (final fix for Nitro SSR)

## Test plan

- [ ] #487 merges cleanly.
- [ ] Merge this. CI matrix runs:
  - all 4 bun versions serverless,
  - all 8 deno versions serverless (1.x unchanged, 2.x uses updated image),
  - 11 SSR frameworks × 3 runtimes.
- [ ] Spot-check image sizes in the auto-generated size report comment.

---

## Local end-to-end validation on Appwrite Sites

On top of the runtimes CI matrix above, I verified both new runtimes end-to-end against a live Appwrite Cloud + Edge instance — cloud via `docker compose -f docker-compose.yml -f docker-compose.edge.yaml`, edge via `kind` + Tilt.

**What I ran**

1. Built `openruntimes/bun:v5-1.3` and `openruntimes/deno:v5-2.6` directly from this branch using the repo's own `ci-runtime-prepare.sh` assembly + `docker build`, then `kind load docker-image --name edge` so the cluster's `imagePullPolicy: IfNotPresent` pods pick them up instead of pulling from the registry.
2. Created two Appwrite Sites from the **same** Astro 4.16 project (`@astrojs/node` adapter, `output: 'server'`). The tarball is byte-identical between them; the only field that differs is `buildRuntime=bun-1.3` vs `buildRuntime=deno-2.6`.
3. Both deployments built, packaged, distributed via edge, and served the rendered HTML with HTTP 200. Edge runtime-pod logs confirm the request was handled by the PR's image (`runtime.image=openruntimes/bun:v5-1.3` / `openruntimes/deno:v5-2.6`) — not a cached static artifact.

**Result**

| Runtime | `buildRuntime` | Build time | Serving | Image |
| --- | --- | --- | --- | --- |
| Bun 1.3 | `bun-1.3` | 91s | bun SSR badge | `openruntimes/bun:v5-1.3` |
| Deno 2.6 | `deno-2.6` | 102s | deno SSR badge | `openruntimes/deno:v5-2.6` |

The landing page reads `process.env.APPWRITE_SITE_RUNTIME_NAME` / `_RUNTIME_VERSION` / `_DEPLOYMENT` at render time, so the runtime badge on the page is evidence the SSR wrapper actually ran — not that a build artifact got served statically. The deployment ID visible on each page matches the deployment created by the Sites API call.

**Bun (`openruntimes/bun:v5-1.3` — JavaScriptCore via bun)**

bun SSR on Appwrite Sites

**Deno (`openruntimes/deno:v5-2.6` — V8 via deno)**

deno SSR on Appwrite Sites

Fixture source + raw screenshots: https://github.com/ChiragAgg5k/pr485-screenshots


runtimes/bun/versions/latest/package.json
```diff
@@ -11,6 +11,8 @@
   "author": "",
   "license": "ISC",
   "dependencies": {
+    "express": "^4.21.2",
+    "srvx": "^0.8.16",
```
Meldiron, line 15: Is this required?

runtimes/bun/versions/latest/src/ssr/injections.ts
Meldiron, line 1: Lets see if agent can now move some of this to javascript, having it shared across runtimes. We might want to also do this for Deno in this same PR- so we have bigger sample of what's JS-common, and what's runtime specific (and just same in Node and Demo).

This aipplies for both injections.ts, and logger.ts.

I assume there will be some differences for sure, but I hope some helpers could be moved to shared file

runtimes/bun/versions/latest/package.json
ChiragAgg5k: Yes — `express` is imported by the shared SSR wrappers at `runtimes/javascript/src/server-<fw>.mjs` (all eight of them use `express.static` for the public/client folder, plus `app.use(handler)`), and `srvx` is imported by `server-tanstack-start.mjs` for `toNodeHandler`. Most frameworks (Astro, SvelteKit, Nuxt, Angular, Analog) do not pull `express` in through the user's own `package.json`, so without the runtime-level fallback at `/usr/local/server/node_modules/` the `bun ./server.mjs` step would fail to resolve the import. Mirrors exactly how node's `runtimes/node/versions/latest/package.json` already declares these two. Happy to drop them if there's a reason to expect users to always declare them themselves, but I don't see one.

runtimes/bun/versions/latest/src/ssr/injections.ts
ChiragAgg5k: Done in two pieces:

**[#487](https://github.com/open-runtimes/open-runtimes/pull/487)** (off `main`, follow-up to the merged #486) moves the node copies into the shared tree:

- `runtimes/javascript/src/ssr/logger.mjs` — unified on `AsyncLocalStorage` (node 16+ and bun both support it natively), dropping `cls-hooked` from node's `package.json`.
- `runtimes/javascript/src/ssr/http-injection.mjs` — `overrideEmit`, `wrapServer`, `wrapCreateServer`, `wrapHttp`. These were byte-identical between node and bun — only the install mechanism differs.

This PR (bottom commit `07a53b93`) then consumes the shared files from the bun side: deletes `src/ssr/logger.ts` and slims `src/ssr/injections.ts` from ~140 LOC to ~50 LOC. What's left bun-specific is only the direct `http.Server.prototype.emit` patch + `Server` / `createServer` wrappers — the part that genuinely differs from node (node uses `import-in-the-middle` / `require-in-the-middle` via `NODE_OPTIONS --import`; bun uses `--preload` and can patch directly).

**Deno:** out of scope here — `runtimes/deno/versions/latest/` has no SSR support today (no framework wrappers, no bundle/env/server helpers, no preload hook, no http injection). Adding Deno sites would be a meaningful new runtime of its own and would need to pick a preload equivalent (Deno's module resolution and http surface differ enough that it's not a cosmetic port). Once Deno SSR lands, it can opt into the same `runtimes/javascript/` overlay via `shared = "javascript"` in the TOML — that part is already wired.

ChiragAgg5k: Added deno scaffolding in `8026ef18` so we now have a three-runtime sample:

**Runtime-specific (each runtime keeps this):**
| File | node | bun | deno |
| --- | --- | --- | --- |
| `helpers/javascript-runner/env.sh` | `OPR_JAVASCRIPT_RUNNER=node` + `NODE_OPTIONS=--import=…` | `OPR_JAVASCRIPT_RUNNER=bun` + `BUN_OPTIONS=--preload=…` | `OPR_JAVASCRIPT_RUNNER="deno run --allow-all --node-modules-dir=auto --import=…"` |
| `src/ssr/injections.{mjs,ts}` | ~25 LOC — `import-in-the-middle` + `require-in-the-middle` registration (node has to hook every module load because `--import` runs too late for CJS requires) | ~40 LOC — direct `http.Server.prototype.emit` patch (preload beats everything else) | ~35 LOC — same direct patch as bun; reads `Deno.env.get(...)` instead of `process.env[...]` |

**JS-common (one copy in `runtimes/javascript/`, overlaid into all three):**
- `src/ssr/system.mjs`
- `src/ssr/logger.mjs`
- `src/ssr/http-injection.mjs` — `overrideEmit` / `wrapServer` / `wrapCreateServer` / `wrapHttp`
- 8 × `src/server-<framework>.mjs` wrappers
- 8 × `helpers/<framework>/bundle.sh`
- 8 × `helpers/<framework>/env.sh`
- 8 × `helpers/<framework>/server.sh` (sources `javascript-runner/env.sh`)

**Deno production gap (explicit follow-ups, not in this PR):**
- Deno's alpine image doesn't currently have npm/pnpm/yarn (the TOML comment already notes this as a TODO). Sites install commands look like `npm install && npm run build && bash /usr/local/server/helpers/astro/bundle.sh` — so deno needs node + npm in its dockerfile before `runtime = { name = "deno", version = "2.6" }` test entries can be added.
- No `deno-sites-*` CI entries wired yet — same deferral as the bun side of this PR.

What's committed scaffolds the deno slot of the shared-JS taxonomy so the pattern is provable three ways; finishing the dockerfile + CI wiring is a follow-up PR once deno's npm support lands.

ci/runtimes.toml
```diff
@@ -261,8 +273,230 @@ output = "./dist/angular"
...
+test = "SSR/NextJS.php"
+output = "./.next"
+
+# next-js-turbopack under deno is intentionally omitted. Turbopack emits a
+# CJS-style `_document.js` without `type: module` that deno's node-compat
+# loader treats as ESM, blowing up with "ReferenceError: require is not
+# defined" at file scope. The non-turbopack next-js_deno variant works fine
+# and is sufficient coverage for deno × Next.js.
+
+[nuxt_deno]
```
Meldiron: I think this is too much combination, we could easily get to 200+ tests

tests/SSR/NextJS.php
```diff
@@ -15,6 +15,11 @@ class NextJS extends SSR
     // server-wrapper and next.config imports, so we fall back to modclean.
     public function testNft(): void
     {
+        $enforcedRuntime = \getenv('ENFORCED_RUNTIME');
+        if ($enforcedRuntime !== false && $enforcedRuntime !== '' && $enforcedRuntime !== 'node') {
+            self::markTestSkipped('@vercel/nft pruning is a node-only cleanup step.');
```
Meldiron: Hmm, when someone switches from node to bun, their cold-start might go from 3s to 15s.

I think we need to support NFT in site runtimes too.

loks0n (approved):

PR #486, opened 2026-04-19
refactor: introduce runtimes/shared/js for JS runtime deduplication

ChiragAgg5k:
## Summary

Introduces a shared tree at `runtimes/shared/js/` for files that are (or would otherwise become) byte-identical across the JS family of runtimes — currently node and bun, with deno a future candidate once it gains SSR. Pure file-location refactor; no behaviour changes.

## Why

Looking at `runtimes/node/versions/latest/` and the in-flight [#485 bun sites support](https://github.com/open-runtimes/open-runtimes/pull/485), 25 files would be literal copies between the two runtimes:

- 8 × `src/server-<framework>.mjs` express wrappers
- `src/ssr/system.mjs` — OPR endpoints + secret auth + safe-timeout helpers
- 8 × `helpers/<framework>/bundle.sh`
- 8 × `helpers/<framework>/env.sh`

Duplicating these into every new JS runtime means every Next.js / Astro / SvelteKit / TanStack tweak lands twice (or worse, lands in one runtime and silently diverges in the other). This PR consolidates them into one tree and teaches the build prep script to overlay that tree onto every JS runtime's build context.

## How

1. **`runtimes/shared/js/`** holds the 25 shared files, mirroring the `src/` and `helpers/` layout that callers already expect.
2. **`ci-runtime-prepare.sh`** gains one `case` block that `cp -R`s `runtimes/shared/js/*` into `./runtimes/.test/` when `$RUNTIME_FOLDER` is `node` or `bun`. It runs *before* the runtime-specific copy, so any runtime can still diverge by shipping a file of the same name under `runtimes/<name>/versions/latest/`.
3. Because the overlay targets the same relative paths the runtime-specific files used, every existing reference keeps working without a change:
   - `cp ../server-astro.mjs ./server.mjs` in `helpers/astro/server.sh` still finds `/usr/local/server/src/server-astro.mjs` — the shared one.
   - `import … from \"./system.mjs\"` in `src/ssr/injections.mjs` still finds `/usr/local/server/src/ssr/system.mjs` — the shared one.
   - `ci/runtimes.toml` install commands (`bash /usr/local/server/helpers/astro/bundle.sh`, `source /usr/local/server/helpers/astro/env.sh`) resolve to the shared copies.

## What stays per-runtime

- `helpers/<framework>/server.sh` — runner (`node` vs `bun`) and preload flag (`NODE_OPTIONS --import` vs `BUN_OPTIONS --preload`) differ.
- `src/ssr/injections.{mjs,ts}` — node uses `import-in-the-middle` / `require-in-the-middle` via `node:module` `register`, bun uses `--preload` + direct `Server.prototype.emit` patching.
- `src/ssr/logger.{mjs,ts}` — node uses `cls-hooked`, bun uses `AsyncLocalStorage`. Could converge later once node drops cls-hooked.
- `src/server.{js,ts}` — functions server (micro vs `Bun.serve`).
- `src/nft.mjs`, `src/nft-nextjs.mjs`, `helpers/prepare-packing.sh` — NFT/modclean pruning is node-only for now.

## Follow-ups

- [Bun sites support (#485)](https://github.com/open-runtimes/open-runtimes/pull/485) will be rebased on top of this branch and rewritten to pick up the shared tree, removing the 25 duplicated files it currently adds.
- Future: converge `logger` once node migrates to `AsyncLocalStorage`.
- Future: port NFT/modclean to bun or factor into shared.

## Test plan

- [ ] `bash tests.sh node-24` with an SSR target (e.g., `astro`, `sveltekit`, `next-js`) — confirm shared overlay produces identical `/usr/local/server/` layout and all existing node SSR tests pass.
- [ ] Inspect `./runtimes/.test/` after `ci-runtime-prepare.sh` for the node runtime and verify every expected file is present at its expected path.
- [ ] CI: node-24 SSR frameworks (astro, sveltekit, next-js, next-js-turbopack, nuxt, tanstack_start_nitro_v2, tanstack_start_nitro_v3, tanstack_start_native, remix, angular, analog) all pass on this branch.


runtimes/javascript/helpers/analog/bundle.sh
Meldiron, line 1: I would just name folder "js", no need for "shared". If you cannot find it in "Node", you look in "JavaScript" - makes sense to me.

So I propose `runtimes/javascript/...`

runtimes/node/versions/latest/helpers/js-runner.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# Per-runtime knobs consumed by the shared SSR server.sh in
+# runtimes/shared/js/helpers/<framework>/server.sh.
+#
+# OPR_JS_RUNNER      — the binary the shared script uses to launch SSR entries
+# NODE_OPTIONS       — ensures user-owned http.createServer flows inherit the
+#                      Open Runtimes contract via src/ssr/injections.mjs
+
+export OPR_JS_RUNNER="node"
```
Meldiron:
```suggestion
export OPR_JAVASCRIPT_RUNNER="node"
```

runtimes/shared/js/helpers/tanstack-start/server.sh
```diff
@@ -6,30 +6,32 @@ shopt -s dotglob
...
+		START_COMMAND="$OPR_JS_RUNNER ./server-http.mjs"
 	fi
 
 	# Realistically never happens - build prevents this
 	if [ -z "$START_COMMAND" ]; then
 		echo 'No server found'
 		exit 1
 	fi
+else
+	START_COMMAND="$OPEN_RUNTIMES_START_COMMAND"
```
Meldiron: Do we need this?

ci-runtime-prepare.sh
```diff
@@ -8,6 +8,14 @@ mkdir -p ./runtimes/.test
 mkdir -p ./runtimes/.test/helpers
 cp -R ./helpers/* ./runtimes/.test/helpers
 
+# Shared JS runtime files (node, bun) — overlaid before runtime-specific so
+# anything in runtimes/<runtime>/versions/... still wins on conflict.
+case "$RUNTIME_FOLDER" in
+node | bun)
```
Meldiron: Let's make a property for it instead, and set it in TOML file for relevant runtimes

runtimes/node/versions/latest/helpers/javascript-runner/env.sh
Meldiron, line 1: I would name this folder `javascript-runner`.

Meldiron (approved):

PR #497, opened 2026-06-08
Add SquashFS build cache support

TorstenDittmann:
## Summary
- install squashfs-tools in the shared runtime base for Alpine and Debian/Ubuntu images
- verify tar, unzip, mksquashfs, and unsquashfs are present in the global tools smoke test
- add helpers/build-cache.sh so executor can invoke build cache restore/save from runtime images
- restore /cache/stores.sqfs into /usr/local/cache/build when available, then run /cache/build-command.sh
- save package manager caches back to /cache/stores.sqfs after successful builds
- treat cache restore/save failures as cache failures, not build command failures

## Testing
- bash -n helpers/build-cache.sh
- confirmed helpers/build-cache.sh contains no apk/apt install commands
- confirmed base-before.dockerfile installs squashfs-tools for both apk and apt-get branches
- bash -n ci_tests.sh
- bash tests.sh node: tools smoke test passed; full suite later failed at unrelated Tests\Serverless\Node::testHeadlessBrowser with 500 vs 200
- focused Dart image build confirmed Debian/Ubuntu branch installs squashfs-tools and exposes mksquashfs/unsquashfs

### DIFF
--- a/base-before.dockerfile
+++ b/base-before.dockerfile
@@ -7,9 +7,9 @@ ENV OPEN_RUNTIMES_HEADERS="{}"
 
 RUN <<EOR
     if [ -f /etc/alpine-release ]; then
-        apk add util-linux zstd
+        apk add util-linux zstd squashfs-tools
     else
-        apt-get update && apt-get install -y util-linux zstd
+        apt-get update && apt-get install -y util-linux zstd squashfs-tools
     fi
 EOR
 
--- a/ci_tests.sh
+++ b/ci_tests.sh
@@ -149,7 +149,7 @@ fi
 
 # Tools test
 echo "Testing tools ..."
-REQUIRED_TOOLS="tar --help && unzip --help"
+REQUIRED_TOOLS="tar --help && unzip --help && command -v mksquashfs && command -v unsquashfs"
 docker run \
 	--platform linux/x86_64 \
 	--name open-runtimes-test-tools \
--- a/ci_tests.sh
+++ b/ci_tests.sh
@@ -149,7 +149,7 @@ fi
 
 # Tools test
 echo "Testing tools ..."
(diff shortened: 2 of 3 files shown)


levivannoort (approved):

TorstenDittmann:
Updated the smoke test to execute both SquashFS binaries with `-help` instead of only checking PATH presence. I avoided `unsquashfs -version` because it prints version info but exits non-zero on Alpine; `unsquashfs -help` exercises the binary and passes on both Alpine and Debian images. Verified with a focused Node Alpine build and Dart Debian image check.

TorstenDittmann:
Re: Greptile note about using `command -v` for SquashFS tools: keeping this intentionally. The tools smoke test is checking expected tool presence in runtime images, not validating distro package integrity. `squashfs-tools` is installed by apk/apt in the shared runtime base, and we trust those package managers to install working binaries. The check is now consistent for all four tools (`tar`, `unzip`, `mksquashfs`, `unsquashfs`) instead of mixing `--help`/`-version` execution checks, since SquashFS help/version exit behavior differs across package versions and made CI brittle.

TorstenDittmann:
Addressed Greptile’s cache-save status concern in `helpers/build-cache.sh`. Cache archive creation now remains direct to `/cache/stores.sqfs`, but it is best-effort: if `mksquashfs` fails after a successful build, the helper logs a warning and still exits with the original build command status.

TorstenDittmann:
Addressed Greptile’s cache-restore concern in `helpers/build-cache.sh`. Cache restore is now best-effort: if `/cache/stores.sqfs` exists but `unsquashfs` fails, the helper logs a warning, resets `CACHE_ROOT`, and continues with a fresh cache so `/cache/build-command.sh` still runs.

PR #498, opened 2026-06-11
feat: native dockerfile approach

levivannoort:
## Why

The build system had grown into ~126 bash scripts: a 5-layer file-copy overlay, 92 near-identical version Dockerfiles, a hardcoded 92-entry publish matrix, and per-runtime lifecycle scripts that had drifted apart. Adding a single runtime version meant editing three places that could fall out of sync.

The goal: make the system docker-native and config-driven, so adding a runtime version is a config-only change and every layer (image assembly, lifecycle, tests, publish) has one source of truth.

## How

- **`ci/runtimes.toml` is the single source of truth.** `bun ci/bake.ts` generates a committed `docker-bake.json`; `bun ci/index.ts` generates the PR test matrix. CI fails if either drifts.
- **Images build with `docker buildx bake`.** One parameterized Dockerfile per family (`ARG BASE_IMAGE` + a few divergence args) replaces all version Dockerfiles. Shared fragments are spliced via `INCLUDE` and embedded as `dockerfile-inline` — no third-party BuildKit frontend. The bash overlay becomes four named build contexts (`helpers` → `shared` → `latest` → `version`).
- **In-container lifecycle is one runner + presence-based hooks.** `helpers/lifecycle/{build,start}.sh` run explicit phases; runtime-specific steps live in `hooks/`. dart/swift/java now use the standard flow (gaining compression + telemetry).
- **Tests run through compose** (`make test ID=node-25`). **Publishing** resolves its job list from `docker buildx bake --print` and pushes with `bake --push`.

**Contracts preserved.** Published tags, the `OPEN_RUNTIMES_*` env contract, executor entrypoints, telemetry keys, and the container/network names the PHP harness hardcodes are unchanged. Verified by building 16/17 families both ways with identical file-tree and image-config.

## Image tagging

We're settling the tagging contract now, as part of this refactor, because publishing is the one layer everything downstream depends on — `appwrite/appwrite` (OSS) and cloud both pull these images, and changing tags later is a breaking, coordinated migration. Doing it here means the new native publish path ships with the final tag scheme baked in.

The mutable tag stays exactly as-is so OSS is unaffected; cloud builds additionally get an immutable, commit-pinned tag (12-char short SHA):

| Build | Tags pushed |
|---|---|
| Release (OSS + cloud) | `openruntimes/node:v5-25` **and** `openruntimes/node:v5-25-a1b2c3d4e5f6` |
| RC | `openruntimes/node:v5-25-rc` |
| Local bake | `openruntimes/node:v5-25` |

Mechanically: a `SHA_SUFFIX` bake variable (default empty) adds a second tag that collapses onto the mutable one for OSS/local/RC builds; only the non-RC publish job sets it from the short SHA.

## Trade-offs

- Family Dockerfiles aren't directly `docker build`-able (`INCLUDE` splicing happens in `ci/bake.ts`); use `docker buildx bake`.
- `docker-bake.json` is generated and committed — must be regenerated after TOML edits (CI gate makes forgetting loud).
- Editing a family Dockerfile affects all its versions at once.
- Bun is now a hard dependency for local test runs and generation.

## Verification

- Image equivalence old-vs-new for 16/17 families.
- Full PHP suites green locally: node-25, python-3.13, go-1.26, java-25, kotlin-1.6, dotnet-8.0/10, ruby-3.0/3.2/4.0, and the next-js SSR suite.
- Recommended before first release: a `workflow_dispatch` `-rc` publish from this branch and a manifest comparison against the previous release.


levivannoort:
## Image-impact review — does this refactor change the published runtime images?

> **Update:** the one behavioral item below (node Next.js packing) has been **resolved** — node now packs Next.js exactly as `main` did. See the table.

**TL;DR:** The native-Dockerfile mechanism (per-runtime `Dockerfile` + `docker-bake` + named build contexts + `helpers/lifecycle` reorg) is **image-neutral** — identical base images, installed packages, `ENTRYPOINT`/`CMD`/`EXPOSE`/`WORKDIR`, server command, and file contents across all **17 runtimes × 92 versions**. A small set of *behavioral* changes ride on the branch (mostly merged feature/fix commits); the only one specific to this refactor (node Next.js packing) has been reverted to match `main`.

### How this was verified
- Static old (`main`) vs new comparison for every runtime/version: base image, packages/pins, build args, lifecycle hooks, baked `src/`, server launch, permissions, platforms.
- `docker-bake.json` confirmed in sync with `ci/runtimes.toml` + the Dockerfiles (92/92 targets, no stale/missing/orphan).
- **Empirically built old (`dockerfile-x`) vs new (`buildx bake`) images for `static-1`, `dart-3.12`, `node-22` and diffed the filesystem, installed packages, env, and metadata.**

### Mechanism = no image impact (empirically confirmed)
The only `/usr/local/server` differences are the intended reorg: `helpers/{prepare-*,before-*,after-build,extract-code,select-compression}.sh` → `helpers/lifecycle/*` + per-runtime `hooks/*` (content-identical renames), `build-cache-*.sh` added, stray `*.dockerfile`/`Dockerfile` no longer copied in. Confirmed **identical**: installed OS packages (incl. node SSR tooling — `alpine-sdk`/`python3-dev`/`py3-pip`/`gcompat`/`bun`), and `CMD`/`ENTRYPOINT`/`EXPOSE`/`WORKDIR`. The only added image env var is `OPEN_RUNTIMES_SERVER_COMMAND` (= the old per-runtime `server.sh` command, consumed by the generic `helpers/server.sh`). Executor contract (`bash helpers/build.sh '<cmd>'`, `helpers/start.sh "bash helpers/server.sh"`) is preserved. `chmod -R +x helpers` is a harmless superset (scripts are sourced/`bash`-invoked, so the exec bit isn't load-bearing).

Highest-risk migrations all verified faithful: dart/swift/java custom lifecycle overrides → global lifecycle + hooks (same compiled artifact, same archive layout); dotnet per-version `prepare-compile.sh` → one `compile.sh` with a `net*` glob (one SDK per image, resolves identically); php multi-stage swoole builder incl. the 8.3 split builder base.

### Behavioral changes that ship in the images
Prod status assessed against edge Helm values (`nft: enabled`, `buildCompression: auto`):

| Change | In prod | Notes |
|---|---|---|
| ✅ **node Next.js packing** — *resolved, now matches `main`* | n/a | node's pre-refactor no-op restored: the shared packer's `modclean` fallback for `.next` is gated behind `OPEN_RUNTIMES_NEXTJS_CLEANUP`, and node sets `=none` (ships `node_modules` untouched). bun/deno keep `modclean` (their prior behavior). Verified by building `node-22`: `CLEANUP=none` on the Next.js path. |
| **static**: in-memory cache **on by default** | live | feature; disable per-deploy with `OPEN_RUNTIMES_STATIC_MEMORY_CACHE_ENABLED=false` |
| **ruby**: `server.rb`/`execute.rb` export-guard + arity fix | live | `fix: ruby 3` |
| **tanstack-start**: new Nitro `node-listener` preset path + wrapper | preset only | additive |
| **auto-compression**: builds ≥10 MB now gzip instead of zstd | live | extractor auto-detects format; functionally safe (size/extract-speed tradeoff) |
| **build cache** wired into `build.sh` (restore/save, `\|\| true`) | live | speeds rebuilds; build output unchanged |

### Sign-off
All changes are mechanism-neutral or intended feature/fix work. The single refactor-introduced behavioral change (node Next.js packing) has been reverted, so published images now pack Next.js exactly as `main` does. No outstanding image-impact concerns.

.github/workflows/version-audit/src/Local.ts
```diff
@@ -41,43 +39,23 @@ export class Local {
...
 
-        const tag = tagMatch[1];
-
-        const key = `${runtime}/${version}`;
-
-        if (!dockerVersions[key]) {
-          dockerVersions[key] = [];
-        }
+        // Edge case due to "-" in name: report under the directory layout name
+        const key =
```
ChiragAgg5k: maybe we should find a better way to handle this edge case?

.github/workflows/publish.yaml
```diff
@@ -16,9 +16,17 @@ jobs:
   prepare:
     runs-on: ubuntu-latest
     outputs:
-      matrix: ${{ steps.matrix.outputs.result }}
+      targets: ${{ steps.matrix.outputs.targets }}
       is_rc: ${{ steps.check.outputs.is_rc }}
     steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
```
ChiragAgg5k: we should use sha hashes

ci/bake.ts
```diff
@@ -0,0 +1,162 @@
...
+// docker-bake.json is committed; together with docker-bake.hcl (variables +
+// _base target) it makes `docker buildx bake <target|group>` work with no
+// generation step.
+
+// @ts-ignore
+import data from './runtimes.toml';
+import { existsSync, readFileSync, writeFileSync } from 'fs';
+import { dirname, join } from 'path';
+
+interface BuildVersion {
```
ChiragAgg5k: maybe a separate types file?

levivannoort:
Fixed the `Tests\Serverless\PHP::testLibrary` failure (`500` instead of `200`).

**Root cause** — not networking. The function uses Guzzle, and the request threw inside the runtime:

```
swoole_curl_setopt(): option[20312] is not supported
  at guzzlehttp/guzzle/src/Handler/CurlFactory.php(637): release()
```

`SWOOLE_HOOK_ALL` enables Swoole's *userland* curl hook (`SWOOLE_HOOK_CURL`), which only maps a fixed set of curl options and throws on the rest. Guzzle 7.x's `CurlFactory::release()` resets `CURLOPT_PREREQFUNCTION` (option `20312`, defined since curl 7.80), which the userland hook doesn't implement → exception → `500`.

**Fix** — one line in `runtimes/php/versions/latest/src/server.php`: switch to the native curl hook, which defers to real curl and supports the full option set while staying coroutine-friendly:

```php
Swoole\Runtime::enableCoroutine($flags = (SWOOLE_HOOK_ALL | SWOOLE_HOOK_NATIVE_CURL) & ~SWOOLE_HOOK_CURL);
```

**This was already failing — it's not a regression from this PR.** The trigger ingredients are byte-identical to `main`:

| | `main` | this branch |
|---|---|---|
| `server.php` hook line | `SWOOLE_HOOK_ALL` | same (only the fix differs) |
| 8.4 base image | `php:8.4.17-cli-alpine3.23` | same |
| Swoole version | `v6.1.1` | same |
| Guzzle constraint | `7.*.*`, no committed lock | same |
| Install command | `composer update` (resolves latest) | same |

The fixture floats `guzzle: 7.*.*` and installs via `composer update`, so every build pulls the newest Guzzle (today 7.13.1). A recent Guzzle 7.x release added the `CURLOPT_PREREQFUNCTION` reset — the day it shipped, this test started failing on **any** build, including `main`. So the docker-bake refactor didn't introduce it; it just surfaced here. The hook fix is the durable resolution (covers all Guzzle versions).

**Verified:** rebuilt php-8.4 and ran the full suite — `OK (39 tests, 349 assertions)`.

> Note: only php-8.4 (Swoole v6.1.1) was rebuilt/verified locally. 8.0 (v4.7.0) and 8.1–8.3 (v5.1.2) share this `server.php`; `SWOOLE_HOOK_NATIVE_CURL` exists in all of them, but those weren't re-run locally.

loks0n (approved):

PR #501, opened 2026-06-16
Split build cache lifecycle scripts

TorstenDittmann:
## Summary
- replace the build-cache wrapper with separate build-cache env, restore, and save helpers
- run cache restore/save best-effort from before-build/after-build lifecycle hooks
- keep cache save atomic and skip cleanly when cache tools, cache root, or artifact directory are missing
- add focused tests for restore/save failure modes, lifecycle wiring, and stale wrapper references

## Tests
- vendor/bin/phpunit tests/BuildCache.php


TorstenDittmann:
Thanks for the review. A few clarifications and follow-ups:

- The cache helpers are intentionally staying in the shared helper layer and cache lifecycle remains globally enabled. Restore/save are best-effort only and never determine build success.
- The unsafe root concern is addressed in `4b2cb481`: `OPEN_RUNTIMES_BUILD_CACHE_ROOT` and `OPEN_RUNTIMES_BUILD_CACHE_ARTIFACT` are no longer configurable. Restore only clears the fixed `/usr/local/cache/build` path, and the artifact path is fixed to `/cache/stores.sqfs`.
- Because the artifact path is fixed to `/cache/stores.sqfs`, the basename artifact case no longer applies.
- The silent `mv` failure concern is addressed in `33af2211`: failed final moves now warn, remove the temporary artifact, and still exit `0` to preserve best-effort cache behavior.
- The Java env inheritance and Dart/Swift overlay observations can result in cache misses for those runtimes, but that is acceptable for this PR. The cache is mainly useful for Node/Bun package manager stores; non-Node runtime cache coverage is not required for correctness and is not blocking this lifecycle split.

Verified after the latest changes with `vendor/bin/phpunit tests/BuildCache.php` and `shfmt`.

loks0n (approved):

PR #512, opened 2026-07-10
perf(images): registry build cache, zstd compression, apk hygiene

levivannoort:
## Why

We run all published runtime images on Kubernetes. Every release currently rebuilds every layer from scratch on ephemeral CI runners with no build cache, so functionally-identical layers get **new digests each publish** — meaning every node re-pulls the whole image on every release even when nothing meaningful changed. This PR targets pull speed and cross-release layer reuse.

## What

- **Registry build cache** (`publish.yaml`, both `deploy` and `deploy-rc`): `cache-to=type=registry,mode=max` + `cache-from`, keyed per target at `openruntimes/buildcache:<target>`. Unchanged instructions now resolve to identical layer digests across releases → nodes pull only changed layers.
- **zstd compression** (`publish.yaml`): push with `compression=zstd,oci-mediatypes=true` for faster decompression on pull. Deliberately **no** `force-compression`, so upstream base layers keep their original gzip digests and stay shareable with other images.
- **apk hygiene** (all runtime Dockerfiles + `docker/base-before.dockerfile`): `apk add --no-cache` everywhere and dropped redundant `apk update`, so the apk index no longer ships inside a layer.
- Regenerated `docker-bake.json` (Dockerfiles are inlined into it; `bun ci/bake.ts --check` passes).

## Not included (deliberately)

**Dependency-first layer reordering** (copy manifest → install → copy source). Blocked from a safe universal change by two constraints in the shared-fragment design: the overlay COPY order (`helpers → shared → latest → version`) is load-bearing for file precedence, and some runtimes read source at build time (e.g. `runtimes/deno/Dockerfile` consumes `src/ssr/deno.json`). Needs per-runtime work; tracked as follow-up.

## Ops note

`openruntimes/buildcache` is created automatically on Docker Hub on first push. If the org restricts auto-creation, create it (private) beforehand.


levivannoort:
Split into focused PRs: #525 (apk hygiene) and #526 (registry build cache + zstd). Closing this in favor of those.

PR #516, opened 2026-07-16
Hide SquashFS build output on success

TorstenDittmann:
## Summary
- capture build-output `mksquashfs` logs during packaging
- replay the logs only when compression fails
- preserve the original failure status

## Testing
- `bash -n helpers/lifecycle/build.sh`
- `shellcheck helpers/lifecycle/build.sh`
- `git diff --check origin/main...HEAD`


TorstenDittmann:
@greptileai review

PR #533, opened 2026-08-05
feat(flutter): inject Appwrite vars into flutter build via dart-defines

jaysomani:


tests/resources/functions/flutter/latest/lib/main.dart
```diff
@@ -54,6 +54,9 @@ class _MyHomePageState extends State<MyHomePage> {
               '$_counter',
               style: Theme.of(context).textTheme.headlineMedium,
             ),
+            const Text(
+              'dart-define:${String.fromEnvironment('TEST_DART_DEFINE')}',
```
Meldiron, line 58: Lets update tests to ensure merging behaves as expected.

As part of dart test, lets include dart define json, and have some variable there like TEST_DART_DEFINE_CUSTOM.

Then, have this show both of them, and then, in assertions, assert both our env var, and their env var (custom) exists

tests/compose.yaml
```diff
@@ -41,6 +41,7 @@ services:
       INSTALL_COMMAND: ${INSTALL_COMMAND}
       OPEN_RUNTIMES_OUTPUT_DIRECTORY: ${OUTPUT_DIRECTORY}
       OPEN_RUNTIMES_ENTRYPOINT: ${ENTRYPOINT}
+      OPEN_RUNTIMES_BUILD_VARS: '{"TEST_DART_DEFINE":"hello_open_runtimes"}'
```
Meldiron: Hmm this seems to be new concept that we will need to send from executor.

Is that true, is that expected by your implementation?

I cant think of a way how we could avoid this, but would love if you explored a bit.

Also I dont like it includes values, not just keys.

Also lets consider if it maybe should be a file with the list istead of env var - which would be better.

runtimes/flutter/versions/latest/hooks/merge_dart_defines.dart
Meldiron, line 1: Lets explore if there is some good simple CLI tool that we could use instead.
In bash we could make JSON out of env vars, im sure a little bit of bash scripting will be enough.
Then to merge two JSONs should also be some simple CLI tool.

Lets see if it can be done easily; if not, dart script might be better

tests/compose.yaml
Meldiron: For exampel I think we have `usr/local/build/.open-runtimes`that could maybe hold this info

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,42 @@
...
+if [ "\$1" = "build" ]; then
+	exec "$REAL_FLUTTER" "\$@" --dart-define-from-file="$OPR_DEFINES_FILE"
+fi
+exec "$REAL_FLUTTER" "\$@"
+EOF
+chmod +x "$OPR_WRAPPER_DIR/flutter"
+
+export PATH="$OPR_WRAPPER_DIR:$PATH"
+
+opr_log "Injecting dart-defines via $OPR_DEFINES_FILE"
```
Meldiron: this seems slightly hacky.

Why dont we always run the prepare - always inject env variables? I would not expect to skip this step.

Also I see some wrapping with "\$@" which is bad practice, it can easily go wrong when developer can customize these params.

Lets explore a bit into what Flutter and Dart actually suggest in their docs. If we find its this, then OK. but really hoping there is clearer solution

tests/compose.yaml
jaysomani: understood you are correct OPEN_RUNTIMES_BUILD_VARS was duplicating secret values into a second place for no reason. will switch to key-value-name only file ✔️

runtimes/flutter/versions/latest/hooks/merge_dart_defines.dart
jaysomani: i was having doubt on this but when exploring i found that jq isn't installed in the Flutter image only bash.

 But helpers/build-manifest.sh already has a jq-with-bash-fallback pattern for exactly this situation, ig 

should we  either add jq to the Flutter Dockerfile (one word — apt-get install -y bash jq) and rewrite in jq, or mirror build-manifest.sh's pure bash approach with no new dependency?

what will be your preference for this or anything which i would have missed should be explored here ?

runtimes/flutter/versions/latest/hooks/build-prepare.sh
jaysomani: Checked Flutter's source  --dart-define-from-file but didn't found any auto-discovery, I think PATH shadowing is unavoidable for custom buildCommands. I checked other CI platform but there approch is different her here they are owning the full build script.

I found a bug when checking the flag goes after "$@" so it incorrectly wins over a user's own --dart-define-from-file. Flutter merges in order with later-wins
i have fixed it will include in next commit
exec "$REAL_FLUTTER" "$1" --dart-define-from-file="$FILE" "${@:2}"

Will also make prepare always run unconditionally as you suggested.

runtimes/flutter/versions/latest/hooks/merge_dart_defines.dart
Meldiron: I would say it's OK to add jq to Flutter Dockerfile. Let's once build without it, once with it, and compare docker image sizes. I would assume difference is very small, matter of few MBs.
If that is the case, we can install jq

tests/CSR/Flutter.php
```diff
@@ -36,4 +37,18 @@ public function testTimings(): void
...
+    public function testDartDefinesInjected(): void
+    {
+        $response = Client::execute(url: '/main.dart.js', method: 'GET');
+        self::assertEquals(200, $response['code']);
+
+        // dart_defines.json fixture wins over the Appwrite-provided
+        // OPEN_RUNTIMES_BUILD_KEYS value on the conflicting TEST_DART_DEFINE key.
+        self::assertStringContainsString('fixture_override', $response['body']);
+        self::assertStringContainsString('custom_value_from_fixture', $response['body']);
+        self::assertStringNotContainsString('hello_open_runtimes', $response['body']);
```
Meldiron, line 52: This setup tests override, conflicts, but never tests for env var that is define by OPR, but not overriden.
We need one more here

tests/compose.yaml
```diff
@@ -41,6 +41,8 @@ services:
       INSTALL_COMMAND: ${INSTALL_COMMAND}
       OPEN_RUNTIMES_OUTPUT_DIRECTORY: ${OUTPUT_DIRECTORY}
       OPEN_RUNTIMES_ENTRYPOINT: ${ENTRYPOINT}
+      TEST_DART_DEFINE: hello_open_runtimes
+      OPEN_RUNTIMES_BUILD_KEYS: TEST_DART_DEFINE
```
Meldiron: Plural sounds like more can be here; lets test for that

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,60 @@
...
+fi
+
+# Merge with the deployment's own dart_defines.json, if present - its values
+# win on conflict. Always (re)write the output so a build with no vars this
+# time doesn't inherit a stale file left over from a previous build.
+if [ -f "dart_defines.json" ]; then
+	jq -s '.[0] * .[1]' <(echo "$OPR_APPWRITE_JSON") "dart_defines.json" >"$OPR_DEFINES_FILE"
+else
+	echo "$OPR_APPWRITE_JSON" >"$OPR_DEFINES_FILE"
+fi
```
Meldiron: I think we dont need to do this.

We talked about flutter command supporting multiple files,

```
flutter --dart-define-from-file=file1.json  --dart-define-from-file=file2.json
```

If that works, we dont need to do merging logic, Flutter will resolve all that for us. We just need to make sure to pass our parameter in correct place (either before user params, or after them)

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,60 @@
...
+
+if ! REAL_FLUTTER="$(command -v flutter)"; then
+	opr_error "flutter not found on PATH, skipping dart-defines injection."
+	return 0
+fi
```
Meldiron: Lets remove this, sounds like useless guard. flutter runtime will always have flutter CLI

Meldiron: Ah I see we need it, we just dont need the if statement

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,60 @@
...
+	OPR_APPWRITE_JSON="$(
+		IFS=',' read -r -a OPR_KEYS <<<"$OPEN_RUNTIMES_BUILD_KEYS"
+		for opr_key in "${OPR_KEYS[@]}"; do
+			opr_key="$(echo "$opr_key" | xargs)"
+			[ -z "$opr_key" ] && continue
+			[[ "$opr_key" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || { opr_error "Skipping invalid env var name: $opr_key" >&2; continue; }
+			jq -n --arg k "$opr_key" --arg v "${!opr_key:-}" '{($k): $v}'
+		done | jq -s 'add // {}'
+	)"
+fi
```
Meldiron: I feel like this could be simpler 🤔

Loop all from OPEN_RUNTIMES_BUILD_KEYS (separated by space). Use JQ to gradually prepare JSON.

If it's done like this to keep it in-memory, its OK by me to write it to file instead. But I think we can simplify inmemory too

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,62 @@
...
+
+cat >"$OPR_WRAPPER_DIR/flutter" <<EOF
+#!/bin/bash
+# Generated by build-prepare.sh
+if [ "\$1" = "build" ]; then
+	# --dart-define-from-file is registered on the platform subcommand (e.g.
+	# "web"), not on "build" itself, so it must come after both — inserting
+	# it between "build" and "web" makes Flutter reject it as unrecognized.
+	# Flutter uses last-wins for repeated --dart-define-from-file flags (verified w/ real builds).
+	# Ours goes first so the deployment's own dart_defines.json (if any) comes after and wins.
```
Meldiron:
```suggestion
```

Lets remove docs from generated file

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,62 @@
...
+	if [ -f "dart_defines.json" ]; then
+		exec "$REAL_FLUTTER" "\$1" "\$2" --dart-define-from-file="$OPR_DEFINES_FILE" --dart-define-from-file="dart_defines.json" "\${@:3}"
+	else
+		exec "$REAL_FLUTTER" "\$1" "\$2" --dart-define-from-file="$OPR_DEFINES_FILE" "\${@:3}"
+	fi
```
Meldiron, line 39:
```suggestion
	exec "$REAL_FLUTTER" "\$1" "\$2" --dart-define-from-file="$OPR_DEFINES_FILE" "\${@:3}"
```

We shouldnt need if/else - if they have JSON file, doesnt nessesarly mean they want to use it. Its up to them to load their JSON files. we just make sure to load ours

runtimes/flutter/versions/latest/hooks/build-prepare.sh
```diff
@@ -0,0 +1,62 @@
...
+	OPR_APPWRITE_JSON="$(
+		read -r -a OPR_KEYS <<<"$OPEN_RUNTIMES_BUILD_KEYS"
+		for opr_key in "${OPR_KEYS[@]}"; do
+			opr_key="$(echo "$opr_key" | xargs)"
+			[ -z "$opr_key" ] && continue
+			[[ "$opr_key" =~ ^[A-Za-z_][A-Za-z0-9_]*$ ]] || { opr_error "Skipping invalid env var name: $opr_key" >&2; continue; }
+			jq -n --arg k "$opr_key" --arg v "${!opr_key:-}" '{($k): $v}'
+		done | jq -s 'add // {}'
+	)"
+fi
```
Meldiron:
```suggestion
# Build a JSON object from the keys listed in OPEN_RUNTIMES_BUILD_KEYS
OPR_APPWRITE_JSON="$(jq -n --arg keys "${OPEN_RUNTIMES_BUILD_KEYS:-}" \
	'[$keys | split(" ")[] | select(. != "")] | map({(.): (env[.] // "")}) | add // {}')"
```

This should work too, much easier to comprehend

Meldiron (approved):

PR #546, opened 2026-08-19
Fix hanging PHP executions after fatal errors

### LINKED ISSUE #545: PHP fatal errors can leave execution requests hanging
## Description

An uncatchable PHP engine fatal such as `E_COMPILE_ERROR` can terminate a Swoole worker while it is processing an execution request. The Swoole server remains available, but the affected request receives no response and waits until the upstream timeout.

For example, loading code with methods that differ only by case triggers this because PHP method names are case-insensitive:

```php
class Broken
{
    public function updateMfa(): void {}
    public function updateMFA(): void {}
}
```

The current `Throwable` handlers do not catch this compile fatal. Execution finalization is skipped, so the response has no status or `x-open-runtimes-log-id`, and the fatal only appears in container output rather than execution errors.

## Reproduction

Against `openruntimes/php:v5-8.0`:

1. Lazily include the broken class from an execution.
2. Send an execution request with logging enabled.
3. The request receives zero bytes and reaches its client timeout.
4. The error log file remains empty.
5. `/__opr/health` continues returning 200 because Swoole replaces the worker.

## Proposed fix

Register one shutdown handler per Swoole worker and maintain a worker-local registry of active request responses and loggers. On fatal shutdown, use `error_get_last()` to:

- write the fatal message to each affected execution error log;
- return `x-open-runtimes-log-id`;
- complete each affected response with HTTP 500;
- allow normal Swoole worker replacement afterward.

Add a PHP-specific regression service so a fatal worker does not interrupt the shared runtime test server.

ChiragAgg5k:
## Summary

- track active PHP execution responses and loggers per Swoole worker
- handle uncatchable PHP fatal shutdowns by writing execution errors and returning HTTP 500 with `x-open-runtimes-log-id`
- add an isolated PHP fatal-error test service and regression fixture

## Why

PHP engine fatals such as `E_COMPILE_ERROR` bypass the existing `Throwable` handlers. The worker exits before finalizing the response, leaving upstream clients waiting until their timeout and putting the error only in container output.

The shutdown handler runs while the worker still owns its active responses. It completes all requests affected by the terminating worker, while normal requests are removed from the registry in `finally`.

Closes #545.

## Tests

- `make format ID=php`
- `bun ci/test.ts php-8.4 --skip-image --skip-formatter`
- `bun ci/test.ts php-8.0 --skip-formatter`

Both PHP 8.4/Swoole 6.1.1 and PHP 8.0/Swoole 4.7.0 completed 40 tests with 355 assertions.

### DIFF
--- a/runtimes/php/versions/latest/src/server.php
+++ b/runtimes/php/versions/latest/src/server.php
@@ -22,6 +22,50 @@ const USER_CODE_PATH = '/usr/local/server/src/function';
 
 $config = new Config();
 $userFunction = null;
+$activeRequests = [];
+
+$fatalShutdownHandler = function () use (&$activeRequests): void {
+    $error = \error_get_last();
+    $fatalTypes = [E_ERROR, E_PARSE, E_CORE_ERROR, E_COMPILE_ERROR, E_USER_ERROR];
+
+    if ($error === null || !\in_array($error['type'], $fatalTypes, true)) {
+        return;
+    }
+
+    $message = \sprintf(
+        '%s in %s:%d',
+        $error['message'],
+        $error['file'],
+        $error['line']
+    );
+
+    // A fatal error terminates the worker, so none of its active requests can continue.
+    foreach ($activeRequests as $request) {
+        $logger = $request['logger'];
+        $response = $request['response'];
+
+        try {
+            $logger->write([$message], Logger::TYPE_ERROR);
+        } catch (\Throwable) {
+        }
+
+        try {
+            $logger->end();
+        } catch (\Throwable) {
+        }
+
+        try {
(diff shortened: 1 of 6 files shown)


runtimes/php/versions/latest/src/server.php
```diff
@@ -22,6 +22,50 @@
 
 $config = new Config();
 $userFunction = null;
+$activeRequests = [];
```
Meldiron, line 25: Slightly worried this is array, what if it was just last request stuff? I would assume last request  killed the server.

Worried to maintain array for memroy issues

Meldiron (approved):

PR #554, opened 2026-08-24
feat(ci): measure server startup time and report it on PRs

ChiragAgg5k:
## What

Measures how long each OPR server takes to become ready and posts the results as a PR comment, alongside the existing image-size report.

- `ci/test.ts` gains a `measureStartup()` step that runs right after `compose up -d`: it polls the primary serve container's `/__opr/health` (authenticated, since the static runtime gates every route behind basic auth) every 50ms and records **cold start** — wall time from the container's `StartedAt` until the first HTTP response. It then reads `/__opr/timings` for the breakdown the lifecycle scripts already record (`extract`, `prepare`, and `startup` — the server-boot time `helpers/lifecycle/start.sh` writes when the ready log line appears) and saves everything to `/tmp/startup-metrics.json`.
- The workflow posts/updates a single **Server Startup Times** comment per PR (same marker + retry/verify pattern as the image-size comment), one row per runtime in the matrix: cold start, extract, prepare, and server boot.

A missing measurement never fails the job — readiness failures are left to phpunit, which waits longer and dumps container logs.

## Example (measured locally)

| Runtime | Cold start | Extract | Prepare | Server boot |
|---------|-----------|---------|---------|-------------|
| `node-latest` | 2.17 s | 1.13 s | 0.00 s | 0.81 s |
| `static-1` | 0.68 s | 0.04 s | 0.01 s | — |

(Static shows — for server boot because it snapshots `timings.txt` into the served directory before the server starts.)

## Test plan

- `bun ci/test.ts static-1` and `bun ci/test.ts node` locally: metrics logged and written to `/tmp/startup-metrics.json`, full phpunit suite passes (the one node failure, `testHeadlessBrowser`, is x86 Chromium crashing under Rosetta emulation on an ARM host — unrelated, passes on CI's native runners).
- The comment step on this PR should populate the table for the runtimes CI selects.

Fixes CLO-4754 (first half — once we have a baseline from CI we can chase the easy wins).

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -178,7 +178,9 @@ jobs:
               if (!fs.existsSync(file)) continue;
               try {
                 fresh[dir.replace(/^startup-metrics-/, '')] = JSON.parse(fs.readFileSync(file, 'utf8'));
-              } catch {}
+              } catch (e) {
+                console.warn(`Skipping unreadable metrics file ${file}: ${e.message}`);
+              }
             }
             if (Object.keys(fresh).length === 0) return;
 
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -102,45 +102,13 @@ jobs:
         timeout-minutes: 15
         working-directory: ./
         run: bun ci/test.ts ${{ matrix.ID }}
-      - name: Report startup time
+      - name: Upload startup metrics
         if: always()
-        uses: actions/github-script@f28e40c7f34bde8b3046d885e986cb6290c5673b # v7
+        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7
         with:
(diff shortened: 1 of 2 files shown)


ChiragAgg5k:
**Review (not blocking)**

CI startup-time metrics after `compose up`: per-runtime artifacts, then one `report-startup-time` job writes a single table. The earlier matrix comment race / unpaginated lookup look fixed.

Nits:
- `measureStartup()` treats any HTTP response as ready (`fetch` does not check `ok`). A 404/502 would stamp a tiny cold start. Count `response.ok` (or at least `status < 500`) before recording `ready`.
- Empty `catch {}` on artifact JSON can drop a runtime row with no log. `console.warn` the path on parse failure.

Hardcoded `test-secret-key` matching compose is fine. Comment table is already filling in.

loks0n (approved):

PR #555, opened 2026-08-24
perf(flutter): preinstall dhttpd at image build time

ChiragAgg5k:
## What

Moves `dart pub global activate dhttpd` from `hooks/server.sh` (runs on every cold start) into the Flutter Dockerfile (runs once at image build).

## Why

The startup baseline from #554 shows Flutter's "server boot" at 0.85–2.68 s while plain Dart runtimes boot in ~0.01 s. The gap is entirely the pub activation the serve hook was doing before starting dhttpd. With dhttpd preinstalled, Flutter cold starts should drop from ~3 s to well under 1 s.

Side effect: the `startup` telemetry for Flutter becomes meaningful — previously the ready line was echoed before the pub install, so the recorded boot time was measuring pub, not the server.

## Test plan

CI's Server Startup Times comment on this PR should show Flutter server boot dropping to near the dhttpd bind time.

## Measured results

Baseline from #554's CI run vs this PR's CI run:

| Runtime | Cold start (before → after) | Server boot (before → after) |
|---------|------------------------------|-------------------------------|
| `flutter-3.24` | 3.396 s → **0.67 s** | 2.68 s → **0 s** |
| `flutter-3.27` | 3.286 s → **0.781 s** | 2.63 s → **0 s** |
| `flutter-3.29` | 2.982 s → **0.531 s** | 2.13 s → **0.01 s** |
| `flutter-3.32` | 3.146 s → **0.689 s** | 2.32 s → **0.01 s** |
| `flutter-3.35` | 2.927 s → **0.766 s** | 2.2 s → **0.01 s** |
| `flutter-3.38` | 2.796 s → **0.835 s** | 1.82 s → **0 s** |
| `flutter-3.41` | 1.918 s → **0.965 s** | 1.09 s → **0.01 s** |
| `flutter-3.44` | 1.463 s → **0.957 s** | 0.85 s → **0 s** |
| `flutter-3.47` | 1.911 s → **0.894 s** | 1.11 s → **0.01 s** |


ChiragAgg5k:
Looks good to me — this is the right place for dhttpd.

Pinning 4.1.0 is correct: 4.2.0+ needs Dart ^3.10, and 3.24 in this matrix is still Dart 3.5. All nine Flutter CI jobs are green and server boot dropped from ~1–3s to 0.00–0.01s, so the binary is on PATH for every image.

One leftover: the ready line still prints before dhttpd binds, so `startup=` is echo time, not bind time. Pre-existing and not blocking; cold start via /__opr/health is the number that actually moved.

Approving.

loks0n (approved):

PR #558, opened 2026-08-24
perf(extract): decompress gzip archives with igzip or pigz when available

ChiragAgg5k:
## What

- `docker/base-before.dockerfile`: installs `isa-l` (igzip) on alpine bases, and `pigz` plus best-effort `isal` on debian/ubuntu bases (igzip is only packaged on bookworm+; older bases keep pigz).
- `helpers/lifecycle/extract.sh`: the gzip branch now prefers `igzip -dc | tar -x`, then `pigz -dc | tar -x`, then the existing `tar -xzf` — same pipefail-guarded subshell pattern the zstd branch already uses.

## Why

Build archives default to gzip, and extraction used tar's built-in zlib — single-threaded and the slowest common decompressor. The baseline from #554 shows extract costing ~1.4 s on node runtimes and 5 s+ on SSR builds (`next-js_deno-2.6`: 5.28 s, `analog`: 11.39 s). igzip decompresses the same streams 2–3× faster with no change to the build side or the archive format, and the edge already decompresses with igzip, so this just brings the runtimes in line.

## Test plan

Full matrix on CI; the startup comment's Extract column should drop across the board, most visibly on node and SSR runtimes. The magic-byte format detection is untouched, so legacy tar/zstd/squashfs archives take the same paths as before.

## Measured results

Baseline from #554's CI run vs this PR's CI run — the fourteen largest extract improvements:

| Runtime | Extract (before → after) | Cold start (before → after) |
|---------|---------------------------|------------------------------|
| `analog` | 11.39 s → **8.12 s** | 12.05 s → **8.676 s** |
| `next-js_deno-2.6` | 5.28 s → **2.44 s** | 6.69 s → **3.807 s** |
| `python-ml-3.13` | 7.24 s → **4.77 s** | 24.149 s → **22.437 s** |
| `python-ml-3.11` | 7.06 s → **4.84 s** | 26.485 s → **22.847 s** |
| `node-24-mjs` | 1.46 s → **0.28 s** | 1.986 s → **0.485 s** |
| `node-26` | 1.47 s → **0.35 s** | 1.772 s → **0.767 s** |
| `node-18.0-mjs` | 1.41 s → **0.29 s** | 2 s → **0.635 s** |
| `node-25` | 1.46 s → **0.34 s** | 1.932 s → **0.791 s** |
| `node-20.0` | 1.49 s → **0.38 s** | 2.12 s → **0.849 s** |
| `node-24` | 1.42 s → **0.35 s** | 1.882 s → **0.868 s** |
| `tanstack_start_nitro_v2` | 1.84 s → **0.79 s** | 3.6 s → **2.103 s** |
| `node-20.0-mjs` | 1.41 s → **0.37 s** | 1.954 s → **0.974 s** |
| `node-23` | 1.36 s → **0.34 s** | 1.997 s → **0.953 s** |
| `node-22-mjs` | 1.34 s → **0.35 s** | 1.887 s → **0.786 s** |


ChiragAgg5k:
Looks good to me.

Shared `base-before` install + `extract.sh` gzip branch is the right place; magic-byte detection is untouched so none/zstd/squashfs/legacy `code.gz` keep their old paths. `igzip -dc | tar -xf -` matches the existing zstd pipefail subshell, and `-dc` is the documented flag pair for both isa-l igzip and pigz.

Alpine 3.16/3.18 (`php-8.0`, `ruby-3.0`, `node-16`) correctly miss `isa-l` (community, 3.19+) and take `pigz` from main — that fallback is load-bearing, glad it landed after the first commit. Debian/Ubuntu package name `isal` (tools, provides `igzip`) vs Alpine `isa-l` is correct; jammy has it in universe, bookworm in main. Size/license are noise (BSD-3 / zlib, ~100KB).

Nits, not blocking:
- The apt heredoc has no `set -e` and ends in `isal || true`, so a missing `pigz` would also drop util-linux/zstd/squashfs-tools from that RUN. Every current apt base has pigz, so latent only.
- `BuildCompression.php` doesn’t lock the igzip/pigz selection the way it does squashfs.

CI: no failures so far on the interesting bases (3.16 pigz, 3.18 pigz, 3.23 igzip, bookworm isal). Startup extract dropped as advertised on node/SSR (`next-js_deno-2.6` 5.28s → 2.44s). I’ll merge once the remaining SSR/swift jobs finish.

loks0n (approved):

PR #559, opened 2026-08-24
fix(java): use jammy bases for JDK 8/11/17 to restore arm64 builds

ChiragAgg5k:
## What

Moves java 8.0/11.0/17.0 from `eclipse-temurin:<v>-jdk-alpine-3.23` to `eclipse-temurin:<v>-jdk-jammy` and updates the comment explaining the constraint.

## Why

The 0.13.25 release workflow failed on `java-8_0/11_0/17_0 (linux/arm64)` with `no match for platform in manifest: not found`: Temurin only publishes multi-arch Alpine images for JDK 21+ — the JDK 8/11/17 Alpine tags are amd64-only (verified against Docker Hub manifests). Jammy is what JDK 18 and 22 already use for exactly this reason. As a result the v5 java 8/11/17 tags on Docker Hub are still the pre-0.13.25 build.

## Test plan

Built all three images locally on an arm64 host with `docker buildx bake java-{8,11,17}_0 --set '*.platform=linux/arm64'` — all succeed (they fail on main with the manifest error). CI runs the java suite; image-size report will show the alpine→jammy size increase, the same tradeoff 18/22 already carry. Needs a patch release after merge to actually publish the fixed images.

### DIFF
--- a/ci/runtimes.toml
+++ b/ci/runtimes.toml
@@ -421,14 +421,15 @@ platforms = ["linux/amd64"]
 
 [java.build.versions]
 "25" = { base = "eclipse-temurin:25-jdk-alpine-3.23", args = { SYSTEM_PACKAGES = "curl" } }
-# Temurin does not publish Alpine 3.23 variants for EOL JDK 18 or 22.
-# The frozen JDK 18 Alpine image is also amd64-only, so keep Jammy for multi-arch builds.
+# Temurin publishes multi-arch Alpine images only for JDK 21+. The Alpine
+# variants of JDK 8/11/17 are amd64-only and JDK 18/22 have no Alpine 3.23
+# build at all, so those five stay on Jammy for linux/arm64 support.
 "22" = { base = "eclipse-temurin:22-jdk-jammy", args = { SYSTEM_PACKAGES = "curl" } }
 "21.0" = { base = "eclipse-temurin:21-jdk-alpine-3.23" }
 "18.0" = { base = "eclipse-temurin:18-jdk-jammy" }
-"17.0" = { base = "eclipse-temurin:17-jdk-alpine-3.23" }
-"11.0" = { base = "eclipse-temurin:11-jdk-alpine-3.23" }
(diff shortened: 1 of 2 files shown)


loks0n (approved):

ChiragAgg5k:
Post-merge notes (not blocking; already shipped in tree).

This is the right fix for the 0.13.25 arm64 miss. Temurin Alpine 8/11/17 are still amd64-only in official-images; Jammy 8/11/17 are amd64+arm64v8. 18/22 were already on Jammy for the same reason; 21/25 correctly stay on Alpine 3.23 (those Alpine tags *are* multi-arch). Catalog (`ci/runtimes.toml`) and generated bake targets moved together. There are no per-version Java Dockerfiles — the shared Dockerfile already has the apk/apt dual path from #528, and version dirs are gradle overlays only. No stale “8/11/17 are Alpine” comments left.

Size (CI vs #558 Alpine baseline, same day):

| runtime | alpine | jammy | delta |
|---|---|---|---|
| java-8.0 | 196.88 MB | 341.56 MB | +145 MB |
| java-11.0 | 305.77 MB | 461.96 MB | +156 MB |
| java-17.0 | 329.57 MB | 476.66 MB | +147 MB |

Matches the tax java-18.0 (520 MB) and java-22 (495 MB) already pay versus alpine java-21.0/25 (356/302 MB). Acceptable in exchange for arm64.

CI: full matrix green, including java-8.0/11.0/17.0. That’s amd64. Arm64 was the actual 0.13.25 failure and is not in PR CI — local `buildx bake java-{8,11,17}_0 --set '*.platform=linux/arm64'` is what closed that gap.

Still needed: a **patch release** so Hub `openruntimes/java:v5-8.0/11.0/17.0` stop serving the pre-0.13.25 images. Until that ships, the merge does not fix production pulls.

Nit only: the catalog comment says “those five stay on Jammy” — 18/22 stayed; 8/11/17 moved back. Cosmetic.

PR #560, opened 2026-08-25
perf(node): cut server boot ~43% by ungating listen and dropping micro

ChiragAgg5k:
> **Walkthrough:** [Unblocking listen()](https://claude.ai/code/artifact/434c03fa-fa7e-481c-b2e1-24025a277edb) — visual explainer: what was blocking the socket, the full micro → `node:http` surface mapping, and how the function contract was verified. The SSR side is in the companion [Boot Path Teardown](https://claude.ai/code/artifact/4c222436-154a-4b5d-bdac-aea75a8da063).

## Why

On edge, the runtime phase we optimize for cold starts is **Server boot** — the `startup=` timing recorded by `helpers/lifecycle/start.sh` between spawning the server command and the `HTTP server successfully started!` line. For Node, our most used runtime, roughly half of that window was spent on work that doesn't need to be on the critical path.

## What was on the boot path

Profiled inside the `node-22` image (median of 20 runs):

| Stage | Cost |
|---|---|
| `require("micro")` (+ `raw-body`, `content-type`, `is-stream`, `arg`) | ~14 ms |
| `import("superjson")` — dynamic ESM import that `server.listen()` was **gated on** via `Logger.ready.then(...)` | ~13 ms |
| logger / config require + listen | ~6 ms |

## Changes

1. **Stop gating `server.listen()` on the superjson import.** The server listens immediately; the ESM loader spin-up no longer blocks boot. `Logger.write()` already falls back to `JSON.stringify` until superjson resolves, which happens within milliseconds of listen — long before any real execution reaches the logger.
2. **Defer the superjson import kickoff to `setImmediate`**, taking even the import initiation cost off the module-load path. `Logger.ready` had no other consumers and is removed.
3. **Replace `micro` with `node:http`.** micro only provided `createServer` wrapping plus `buffer()`/`send()` — reimplemented inline (~70 lines) with the same behavior: the 20 MB body limit surfaces as an error to the existing catch (oversized bodies still answer with a response instead of a connection reset), and Buffer, stream, object and number bodies are handled exactly as micro's `send()` did. Drops `micro` and 10 transitive packages from the image.

`NODE_COMPILE_CACHE` warmed at image build was also tested and showed **no measurable win** — the boot cost here is module resolution I/O, not V8 compile — so it was left out.

## Timings

Node process boot (spawn → "HTTP server successfully started", `process.uptime()`-based, n=20 each, arm64; production amd64 vCPUs scale the absolute savings up):

| Variant | Median | Δ |
|---|---|---|
| baseline | 47.5 ms | — |
| ungate listen from superjson | 38.6 ms | −19% |
| + node:http instead of micro | 31.4 ms | −34% |
| + defer superjson kickoff | **26.9 ms** | **−43%** |

End-to-end `startup=` as recorded by the lifecycle telemetry (includes the bash wrapper overhead, 10 ms resolution): median **60 ms → 40 ms**.

## Testing

- Full `node-22` suite: 39/40 passing — the one failure, `testHeadlessBrowser`, reproduces identically on `main` when run on arm64 (x86-only chromium binary) and is unrelated.
- The first run caught a real behavior difference (`res.send(50)` number bodies) which is covered by the existing `testDeprecatedMethodsUntypedBody` and now passes.
- Manually verified health endpoint, JSON execution, timings endpoint, and that oversized (23 MB) bodies still receive a 500/413 response.

## Server Startup Times (CI)

"Server boot" from this PR's CI [Server Startup Times](https://github.com/open-runtimes/open-runtimes/pull/560#issuecomment-5404303970) comment vs the same rows measured without this PR (from #561's run, which is `main` for the function server — same CI harness, amd64 runners):

| Runtime | Server boot (without this PR) | Server boot (this PR) | Δ |
|---|---|---|---|
| `node-20.0` | 0.15 s | 0.09 s | −40% |
| `node-22` | 0.16 s | 0.08 s | −50% |
| `node-22-mjs` | 0.17 s | 0.05 s | −71% |
| `node-24` | 0.13 s | 0.05 s | −62% |
| `node-25` | 0.10 s | 0.06 s | −40% |
| `node-26` | 0.13 s | 0.06 s | −54% |
| `node-16.0` – `node-19.0` (legacy, gated listen) | 0.17–0.21 s | 0.15–0.19 s | ~unchanged by design |

Legacy Nodes keep the superjson gate (no `require(esm)`), so their boot is intentionally unchanged; every Node with `require(esm)` support boots roughly twice as fast.

### DIFF
--- a/runtimes/node/versions/latest/package-lock.json
+++ b/runtimes/node/versions/latest/package-lock.json
@@ -12,7 +12,6 @@
         "@vercel/nft": "^0.29.2",
         "express": "^4.21.2",
         "import-in-the-middle": "^1.14.4",
-        "micro": "^9.3.4",
         "require-in-the-middle": "^8.0.0",
         "srvx": "^0.8.16",
         "superjson": "^2.2.6"
@@ -220,11 +219,6 @@
         "url": "https://github.com/chalk/ansi-styles?sponsor=1"
       }
     },
-    "node_modules/arg": {
-      "version": "4.1.0",
-      "resolved": "https://registry.npmjs.org/arg/-/arg-4.1.0.tgz",
-      "integrity": "sha512-ZWc51jO3qegGkVh8Hwpv636EkbesNV5ZNQPCtRa+0qytRYPEs9IYT9qITY9buezqUH5uqyzlWLcufrzU2rffdg=="
-    },
     "node_modules/array-flatten": {
       "version": "1.1.1",
       "resolved": "https://registry.npmjs.org/array-flatten/-/array-flatten-1.1.1.tgz",
@@ -376,14 +370,6 @@
         "balanced-match": "^1.0.0"
       }
     },
-    "node_modules/bytes": {
-      "version": "3.0.0",
-      "resolved": "https://registry.npmjs.org/bytes/-/bytes-3.0.0.tgz",
-      "integrity": "sha1-0ygVQE1olpn4Wk6k+odV3ROpYEg=",
-      "engines": {
-        "node": ">= 0.8"
-      }
-    },
     "node_modules/call-bind-apply-helpers": {
       "version": "1.0.2",
       "resolved": "https://registry.npmjs.org/call-bind-apply-helpers/-/call-bind-apply-helpers-1.0.2.tgz",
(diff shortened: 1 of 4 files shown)


ChiragAgg5k:
Looked at HEAD (888246af), not the original patch.

The two Greptile races are fixed: require(esm) nodes listen immediately and load superjson synchronously on the first structured log; 16–21 still gate listen on import("superjson"), same as main. Health/timings don't touch the logger. Number bodies and the 20MB limit match what the suite already asserts (500|413 on 23MB, "50" for untyped send).

micro is only used from this server.js; SSR jobs are green.

One ask: please update the PR body — it still describes setImmediate + always-ungate + JSON.stringify fallback, which is not what's in HEAD. The 413 on buffer() is also unused (inner catch always sends 500, as with micro). Neither should block merge.

HarshMN2345 (approved):

PR #569, opened 2026-09-11
[CI verification only] feat: add Jaspr SSR support for Flutter runtime (#563)

Meldiron:
**Do not merge — this is a CI-only mirror of #563.**

Same 3 commits as #563 by @jaysomani, rebased onto current `main` and pushed to an internal branch so the workflow runs with a writable `GITHUB_TOKEN`. Diff is identical: 17 files, +302.

**Why #563 looks red:** `.github/workflows/test.yaml` runs `on: [pull_request]`, and the final "Docker Image Sizes" step calls `github.rest.issues.createComment` (line 144). On fork PRs that token is read-only, so the step throws `HttpError: Resource not accessible by integration` (403) and fails **every** matrix job — after the tests have already run. It is unrelated to the contributor's code.

Once this run is green, #563 gets merged and this PR is closed.


Meldiron:
Closing — verification done, results carried over to #563.

Confirmed on #563's own run (fork PR, 113 jobs green where previously zero matrix jobs could pass):

- **Fork PRs always red** — `continue-on-error` on the two comment steps. `Report image size` now succeeds on a fork PR instead of 403ing.
- **EOL Debian apt** — all 22 `dart-*` jobs green, including `dart-2.16`–`dart-3.0` which were failing on expired `bullseye-security` metadata.
- **Jaspr helper never shipped** — moved under `versions/latest/helpers/`; the server starts and 19 tests now run (was 19 errors at startup).

Remaining on #563: Jaspr implements none of the open-runtimes HTTP contract — no `x-open-runtimes-secret` enforcement, no `/__opr/health`, no `/__opr/timings`. `testWrongSecret`/`testEmptySecret` return 200 instead of 401/500. Needs a design decision, tracked on #563.

